PR #2, opened 2022-08-01
V1

lifeislikethisonly:
I have added partial implementation of the program wrappers.

### DIFF
--- a/lib/program/Block.h
+++ b/lib/program/Block.h
@@ -4,6 +4,7 @@
 #include "Instruction.h"
 #include "LLVMtoVanguard.h"
 #include <unordered_set>
+#include <list>
 #include "llvm/IR/BasicBlock.h"
 
 namespace vanguard{
@@ -13,7 +14,7 @@ namespace vanguard{
 
         const Function* getFunction();
 
-        llvm::SymbolTableList< llvm::Instruction >* getInstructionsList();
+        std::list<Instruction *> getInstructionsList();
 
         bool isEntryBlock();
 
--- a/lib/program/Block.h
+++ b/lib/program/Block.h
@@ -1,0 +1,28 @@
+#ifndef VANGUARD_PROGRAM_BLOCK_H
+#define VANGUARD_PROGRAM_BLOCK_H
+
+#include "Instruction.h"
+#include "LLVMtoVanguard.h"
+#include <unordered_set>
+#include "llvm/IR/BasicBlock.h"
+
+namespace vanguard{
+    class Block{
+    public:
+        Block(llvm::BasicBlock &block);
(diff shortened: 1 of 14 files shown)


lib/program/LLVMtoVanguard.h
```diff
@@ -0,0 +1,41 @@
...
+namespace vanguard{
+    class Module;
+    class Function;
+    class Instruction;
+    class Block;
+
+    class LLVMtoVanguard{
+
+        public:
+            static vanguard::LLVMtoVanguard* getSingletonLLVMtoVanguard();
```
stephensj2, line 20: Please rename this to getInstance

lib/program/LLVMtoVanguard.h
```diff
@@ -0,0 +1,41 @@
...
+        
+            Module* getVanguardModule(llvm::Module* module);
```
stephensj2, line 22: Rename each of these translate

lib/program/Block.h
```diff
@@ -0,0 +1,28 @@
...
+#include "llvm/IR/BasicBlock.h"
+
+namespace vanguard{
+    class Block{
+    public:
+        Block(llvm::BasicBlock &block);
+
+        const Function* getFunction();
+
+        llvm::SymbolTableList< llvm::Instruction >* getInstructionsList();
```
stephensj2, line 16: This should be able to return a list of Vanguard Instructions

lib/program/Function.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+    std::string Function::getName(){
+        return std::string(function.getName());
+    }
+
+    bool Function::getParams(){
+        return false;
+    }
+
+    llvm::FunctionType* Function::getFunctionType(){
```
stephensj2, line 19: I think we can probably delete this function. It looks like we'll likely need to use it internally but shouldn't be exposed to the user.

lib/program/LLVMtoVanguard.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+    LLVMtoVanguard* LLVMtoVanguard::getSingletonLLVMtoVanguard(){
+        if (singletonLLVMtoVanguard == 0){
+            singletonLLVMtoVanguard = new LLVMtoVanguard();
+        }
+        return singletonLLVMtoVanguard;
+    }
+
+    vanguard::Module* LLVMtoVanguard::getVanguardModule(llvm::Module* module){
+        return moduleMap[module];
```
stephensj2, line 22: These will always return nullptr.

stephensj2 (approved):

PR #10, opened 2022-08-20
Multi-Module Analysis

stephensj2:
This PR moves Vanguard from a library that uses opt to its own executable. This is done so we can analyze multiple modules at once rather than one at a time.


lib/Vanguard.cpp
```diff
@@ -1,51 +1,203 @@
...
+    llvm::initializeEntryExitInstrumenterPass(registry);
+    llvm::initializePostInlineEntryExitInstrumenterPass(registry);
+    llvm::initializeUnreachableBlockElimLegacyPassPass(registry);
+    llvm::initializeExpandReductionsPass(registry);
+    llvm::initializeExpandVectorPredicationPass(registry);
+    llvm::initializeWasmEHPreparePass(registry);
+    llvm::initializeWriteBitcodePassPass(registry);
+    llvm::initializeHardwareLoopsPass(registry);
+    llvm::initializeTypePromotionPass(registry);
+    llvm::initializeReplaceWithVeclibLegacyPass(registry);
```
kferles, line 87: Maybe we can create a method for all initializations.

kferles (approved):
I think this makes sense to me. I added a minor comment, but I didn't spend much time reviewing the CMake stuff.

lib/detectors/DetectorRegistry.h
```diff
@@ -0,0 +1,29 @@
...
+    public:
+        ~DetectorRegistry();
+        static DetectorRegistry &getInstance();
+        bool add(const std::string& name, ProgramDetector *detect);
+        ProgramDetector *get(const std::string& name);
+        std::vector<ProgramDetector *> all();
+
+    private:
+        static DetectorRegistry *instance;
+        std::unordered_map<std::string, ProgramDetector *> registry;
```
Technius, line 22: nitpick: it might be better to use `std::unique_ptr<ProgramDetector>` to avoid having to deal with manual new and delete.

lib/detectors/FunctionDetector.cpp
```diff
@@ -0,0 +1,32 @@
...
+// Created by Jon Stephens on 8/13/22.
+//
+
+#include "FunctionDetector.h"
+
+namespace vanguard {
+
+    void FunctionDetector::detect(CompilationUnit &unit) {
+        std::unordered_set<Function *> processed;
+        std::list<Function *> fnWorklist;
```
Technius, line 11: nitpick: I think `std::deque` is better for this use case as it makes fewer heap allocations.

lib/detectors/FunctionDetector.cpp
```diff
@@ -0,0 +1,32 @@
+//
+// Created by Jon Stephens on 8/13/22.
+//
+
+#include "FunctionDetector.h"
+
+namespace vanguard {
+
+    void FunctionDetector::detect(CompilationUnit &unit) {
+        std::unordered_set<Function *> processed;
```
Technius, line 10: nitpick: might be better to use `llvm::SmallPtrSet`.

lib/detectors/FunctionDetector.cpp
```diff
@@ -0,0 +1,32 @@
...
+        for(auto f : unit.getAllFunctions()) {
+            fnWorklist.push_back(f);
+        }
+
+        while(!fnWorklist.empty()) {
+            Function *curFn = fnWorklist.front();
+            fnWorklist.pop_front();
+            processed.insert(curFn);
+
+            bool modified = detect(*curFn);
```
Technius, line 21: note: for debugging purposes, we might want to bound the number of iterations in case a bug causes the iteration to diverge.

lib/Vanguard.cpp
```diff
@@ -1,51 +1,203 @@
...
+
+    auto &llvmTrans = vanguard::LLVMtoVanguard::getInstance();
+
+    std::vector<std::unique_ptr<llvm::Module>> modules;
+    std::vector<vanguard::CompilationUnit *> units;
+    for (auto inFile: inputFiles) {
+        auto module = parseIRFile(inputFiles[0], Err, ctxt, setDataLayout);
+        if (!module) {
+            Err.print(argv[0], llvm::errs());
+            return 1;
```
Technius, line 164: super minor nitpick:
```suggestion
            return EXIT_FAILURE;
```

lib/CMakeLists.txt
```diff
@@ -56,25 +60,58 @@ add_subdirectory(program)
...
+        LLVMSupport
+        LLVMTarget
+        LLVMTransformUtils
+        LLVMVectorize
+        LLVMPasses
+        )
+
+target_link_libraries(Vanguard PUBLIC
+        ${LLVM_LINK_COMPONENTS}
+        )
```
Technius, line 115: To help with testing, I would split this into `libVanguard` (library only) and `Vanguard` (executable only). Then we can reuse the `libVanguard` target when we set up the dependencies for unit tests.

lib/Vanguard.cpp
Technius: Also, do we really need all of these?

lib/Vanguard.cpp
```diff
@@ -1,51 +1,203 @@
...
+    FAM.registerPass([&] { return std::move(AA); });
+
+    // Register all the basic analyses with the managers.
+    PB.registerModuleAnalyses(MAM);
+    PB.registerCGSCCAnalyses(CGAM);
+    PB.registerFunctionAnalyses(FAM);
+    PB.registerLoopAnalyses(LAM);
+    PB.crossRegisterProxies(LAM, FAM, CGAM, MAM);
+
+    llvm::ModulePassManager MPM;
```
Technius, line 142: MPM currently doesn't have any passes. Are we going to do some preprocessing in the future?

lib/Vanguard.cpp
```diff
@@ -1,51 +1,203 @@
...
+    vanguard::Program prog(units);
+
+    if(detectors.empty()) {
+        for(auto detector : detectorRegistry.all()) {
+            performDetection(detector, prog);
+        }
     }
-}
+    else {
+        for(auto name : detectors) {
```
Technius, line 181: This loop should be split into two: 1) validate the detectors; 2) run the detectors.

Currently, if a detector is unknown, this loop will print out a warning that will get mixed up with the (valid) detector results.

Technius (commented):
Overall structure looks good to me. Some nitpicks below.

PR #13, opened 2022-08-26
added files needed for ci

momo-ai:
this only works for libVanguard


Technius (approved):
Maybe it needs a review before it runs?

momo-ai:
didnt work

PR #19, opened 2022-09-02
Enable build caching with cachix

Technius:


momo-ai (approved):
All seems good from my limited PoV. My understanding is ci.yml now pulls deps from the veridise-public cache, libVanguard.nix was changed to filter out nix files, flake.lock, and the deps folder.

PR #21, opened 2022-09-15
Fixing a bug and making factories actually singletons

stephensj2:
When making changes for Kostas I forgot to make the factories actually singleton classes. Caused a link error when he tried using the getInstance static function.

### DIFF
--- a/lib/domain/libBlockchain/BlockchainFactory.h
+++ b/lib/domain/libBlockchain/BlockchainFactory.h
@@ -12,17 +12,15 @@
 #include "BlockchainModel.h"
 
 namespace vanguard {
-    class BlockchainFactory : public LLVMFactory {
+    class BlockchainFactory : public UnitFactory {
     public:
-        static UnitFactory *getInstance() {
+        static BlockchainFactory *getInstance() {
             if(instance == nullptr) {
-                instance = new LLVMFactory();
+                instance = new BlockchainFactory();
             }
 
-            return instance;
+            return dynamic_cast<BlockchainFactory *>(instance);
         }
-
-        BlockchainFactory() = default;
 
         BlkType<UnknownType> *createBasicType(const llvm::Module &module, BlockchainModel &model, std::string name);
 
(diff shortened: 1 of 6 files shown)


lib/domain/libBlockchain/BlockchainFactory.h
```diff
@@ -12,18 +12,16 @@
...
     public:
-        static UnitFactory *getInstance() {
+        static BlockchainFactory *getInstance() {
             if(instance == nullptr) {
-                instance = new LLVMFactory();
+                instance = new BlockchainFactory();
             }
 
-            return instance;
+            return dynamic_cast<BlockchainFactory *>(instance);
```
kferles, line 22: The dynamic cast here is unnecessary.

lib/program/LLVMFactory.h
```diff
@@ -15,16 +15,14 @@ namespace vanguard{
...
     class LLVMFactory : public UnitFactory {
         public:
-            static UnitFactory *getInstance() {
+            static LLVMFactory *getInstance() {
                 if(instance == nullptr) {
                     instance = new LLVMFactory();
                 }
 
-                return instance;
+                return dynamic_cast<LLVMFactory *>(instance);
```
kferles, line 23: Same as above.

kferles (changes requested):

lib/domain/libBlockchain/BlockchainFactory.h
stephensj2: The dynamic cast is necessary because the type of instance is UnitFactory and here we are returning a BlockchainFactory

kferles: But we know locally that instance is of type BlockChainFactory, you can simply do `(BlockchainFactory *) instance`, which avoids the runtime check.

stephensj2: If someone interacts with the factories incorrectly (i.e. both BlockchainFactory and LLVMFactory getInstance are called) that could lead to errors that are more difficult to diagnose than accessing null.

kferles: Reviewing the API again, it seems there is a need for a dynamic cast if we want to throw an exception when clients call the wrong getInstance method. If this is the case, maybe it's better to remove the getInstace from all subclasses of UnitFactory and replace them with an init function.

kferles: @stephensj2: I saw your comment after I wrote mine. Yes, I agree with that. Do you want to refactor the API or leave it as is?

kferles (approved):

PR #17, opened 2022-08-31
nix + CI  

momo-ai:
Nix bundles for solidity preprocessor, rust preprocessor, libVanguard, Solang, solc-typed-ast build
TODO: rewrite the filepaths in run.py to the nix dependencies

PR #23, opened 2022-09-16
Add some dependencies to nix build

Technius:
Cleaned up version of #17 rebased onto latest changes in v1.


momo-ai (approved):
Once we have won the hearts of the masses, we should just replace the deps folder w/ nix and consolidate all the .nix files there

PR #31, opened 2022-10-11
Fix C++ driver bugs

Technius:
* Fix segfault when loading Solidity summary
* Fix detector registration not working (no detectors returned, even with all flags provided)
* Add C++ developer environment settings to flake.nix


kferles (approved):
I will trust you with the build stuff, as I don't know much about nix. The C++ stuff LGTM.

PR #32, opened 2022-10-11
New driver script

Technius:
This adds a new Vanguard driver script `vanguard_driver.py` that is a cleaned up and improved copy of `run.py`.

* The new driver script assumes all executables are on `PATH`.
* `run.py` is deprecated and should be removed, but it is left in there to avoid breaking people's workflows for now.
* This will have conflicts with #30.
* Only Solang is supported right now. Rust support has been temporarily disabled since it is complex to set up.

### DIFF
--- a/lib/Vanguard.cpp
+++ b/lib/Vanguard.cpp
@@ -2,6 +2,7 @@
 // Created by Jon Stephens on 8/19/22.
 //
 
+#include <llvm/ADT/STLExtras.h>
 #include <llvm/IR/LLVMContext.h>
 #include <llvm/Support/InitLLVM.h>
 #include <llvm/Support/CommandLine.h>
@@ -19,6 +20,7 @@
 #include <llvm/MC/MCTargetOptions.h>
 #include <llvm/Passes/StandardInstrumentations.h>
 #include <llvm/Support/TargetRegistry.h>
+#include <llvm/Support/FileSystem.h>
 #include <llvm/CodeGen/CommandFlags.h>
 #include <llvm/CodeGen/TargetPassConfig.h>
 #include <llvm/Analysis/AliasAnalysis.h>
@@ -26,6 +28,7 @@
 #include "program/Base.h"
 #include "detectors/DetectorRegistry.h"
 #include "program/Factory.h"
+#include <llvm/Support/FormatVariadic.h>
 //#include "domain/libBlockchain/BlockchainFactory.h"
 //#include "domain/libBlockchain/Blockchain.h"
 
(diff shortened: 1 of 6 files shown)


lib/Vanguard.cpp
```diff
@@ -26,11 +27,15 @@
 #include "program/Base.h"
 #include "detectors/DetectorRegistry.h"
 #include "program/Factory.h"
+#include <llvm/Support/FormatVariadic.h>
 //#include "domain/libBlockchain/BlockchainFactory.h"
 //#include "domain/libBlockchain/Blockchain.h"
 
 static llvm::cl::list<std::string> detectors("detectors", llvm::cl::desc("Vanguard Detectors to Run"), llvm::cl::CommaSeparated, llvm::cl::OneOrMore, llvm::cl::Optional);
 static llvm::cl::list<std::string> inputFiles(llvm::cl::Positional, llvm::cl::desc("<Input files>"), llvm::cl::OneOrMore);
+static llvm::cl::list<std::string>
```
kferles, line 36: We removed the summary file command line option a while back. We now have the convention that for every LLVM file, let's say file.ll or file.bc), there exists a file.json in the same directory. Probably the solang front-end hasn't been updated to generate this file structure, because we haven't used this front-end lately.

lib/Vanguard.cpp
```diff
@@ -206,14 +211,69 @@ int main(int argc, char **argv) {
...
+          return EXIT_FAILURE;
+        }
+
         auto &factory = vanguard::BlockchainDomain::Factory::instance();
         std::vector<vanguard::BlockchainDomain::CompilationUnit *> units;
         units.reserve(modules.size());
         for(auto &mod : modules) {
             std::string filename = filenames[mod.get()];
-            std::string summary = filename.substr(0, filename.length() - 2) + "json";
+
```
kferles, line 270: This will break other test cases we have, especially the ones for near.

lib/Vanguard.cpp
Technius: Makes sense, I'll remove the flag and instead check that the summary files exist.

lib/Vanguard.cpp
Technius: Will revert to the convention you mentioned.

kferles (approved):
LGTM

lib/Vanguard.cpp
Technius: After testing it out some more, I'm a bit confused about how the files are supposed to be laid out.

Based on my reading of the Rust part for the current script, the files are laid out as follows:
```plain
src/my_src.rs

out_dir/
  my_src.ll                 # generated by cargo/rustc
  my_src_deploy.bc          # converted from .ll by llvm-as
  my_src_summary.json       # created by rust summarizer
```
In `Vanguard.cpp`, the summary is assumed to be located at `my_src_deploy.json`. So either the script must be wrong or `Vanguard.cpp` must be wrong. The easiest change we can make is to change the driver script to adjust to `Vanguard.cpp`. Thoughts?

PR #33, opened 2022-10-11
Package SVF in Nix

Technius:
Adds SVF (for use with the svf branch) to the Nix packaging.

### DIFF
--- a/nix/svf/compile-flags.patch
+++ b/nix/svf/compile-flags.patch
@@ -1,0 +1,13 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 2ce7c56d..aa97c485 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -20,7 +20,7 @@ if(NOT COMMAND add_llvm_library)
+         # add -std=gnu++14
+         set(CMAKE_CXX_EXTENSIONS ON)
+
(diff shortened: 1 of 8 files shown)


nix/svf/compile-flags.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 2ce7c56d..aa97c485 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -20,7 +20,7 @@ if(NOT COMMAND add_llvm_library)
+         # add -std=gnu++14
+         set(CMAKE_CXX_EXTENSIONS ON)
+
+-        set(COMMON_FLAGS "-fPIC -Werror -Wall")
++        set(COMMON_FLAGS "-fPIC")
```
Technius, line 10: Based on the build failure, probably need to add `-Wformat` or `-Wall`.

nix/svf/default.nix
```diff
@@ -0,0 +1,42 @@
...
+
+    # We cannot use the normal z3 from nixpkgs, because it uses the old
+    # ./configure based build, which does not have pkg-config files or CMake
+    # config files.
+    z3-cmake
+  ];
+
+  nativeBuildInputs = [ cmake ninja ];
+
+  cmakeFlags = ["-DCMAKE_CXX_FLAGS="];
```
Technius, line 38: This should probably be removed

PR #34, opened 2022-10-13
kferles/add svf

kferles:
Initial integration with SVF framework:
* Runs SVF analyses in a demand-driven way.
* Reimplementation of the NEAR callback detector using SVF dependence analysis.
* Build processes now also build SVF as a dependency.
* Updates README.md with new instructions on how to run and build Vanguard from source.


.gitmodules
```diff
@@ -4,3 +4,6 @@
 [submodule "deps/RustPreprocessor"]
 	path = deps/RustPreprocessor
 	url = https://github.com/Veridise/RustPreprocessor.git
+[submodule "SVF"]
+	path = SVF
+	url = git@github.com:SVF-tools/SVF.git
```
Technius, line 9: GHA isn't able to clone because this is a ssh url, should change to https

CMakeLists.txt
```diff
@@ -21,6 +21,60 @@ set(CMAKE_CXX_STANDARD_REQUIRED on)
...
+    else()
+        set(SVF_LIB "${SVF_DIR}/Release-build/lib/libSvf.a")
+    endif()
+else()
+    MESSAGE (STATUS "building SVF in release mode")
+    set(SVF_LIB "${SVF_DIR}/Release-build/lib/libSvf.a")
+endif()
+set(SVF_HEADER "${SVF_DIR}/include")
+include_directories("${SVF_HEADER}")
+MESSAGE("${SVF_HEADER}")
```
Technius, line 40: We'll probably want to move this code to a `FindSVF.cmake` in the future.

.gitmodules
kferles: Since you set up SVF with nix, we no longer need SVF as a submodule, right?

CMakeLists.txt
kferles: OK, will do that.

Technius: nvm, I just patched SVF to generate the CMake config file instead.

.gitmodules
Technius: Yeah. But this means that building outside of Nix will no longer work, unless we also provide the instructions for it.

Technius:
Branch with the SVF + nix setup, integrated into our current CMake configuration: https://github.com/Veridise/Vanguard/tree/add-svf-with-nix

.gitmodules
kferles: Yes, to me SVF should be handled in the same way as LLVM. They are both dependencies, not part of Vanguard. I will update the instructions in the README before merging.

lib/analysis/interproc-analysis/SVFUtils.h
```diff
@@ -0,0 +1,159 @@
...
+    class SVFUtils {
+    public:
+
+        static bool dependsOnTrg(const llvm::Value *src, const llvm::Value *trg) {
+            if (cachedDeps.find(src) == cachedDeps.end()) {
+                std::set<const SVF::VFGNode*> deps;
+                return traverseMemDependencies(src, deps, trg);
+            }
+            else {
+                auto deps = cachedDeps[src];
```
Technius, line 29: nit: I'd make this explicitly a reference
```suggestion
                auto &deps = cachedDeps[src];
```

lib/analysis/interproc-analysis/SVFUtils.h
```diff
@@ -0,0 +1,159 @@
...
+#include <WPA/Andersen.h>
+#include <SVF-FE/SVFIRBuilder.h>
+#include <Graphs/SVFGNode.h>
+
+namespace analysis {
+
+    class SVFUtils {
+    public:
+
+        static bool dependsOnTrg(const llvm::Value *src, const llvm::Value *trg) {
```
Technius, line 23: nit: I think this definition can go in the cpp file

lib/analysis/interproc-analysis/SVFUtils.h
```diff
@@ -0,0 +1,159 @@
...
+                std::set<const SVF::VFGNode*> deps;
+                return traverseMemDependencies(src, deps, trg);
+            }
+            else {
+                auto deps = cachedDeps[src];
+                return std::any_of(deps.begin(), deps.end(), [&trg](auto node) { return node->getValue() == trg; });
+            }
+        }
+
+        static std::vector<const llvm::Value*> valueMemDeps(const llvm::Value *src) {
```
Technius, line 34: nit: I think this definition can go in the cpp file

lib/analysis/interproc-analysis/SVFUtils.h
```diff
@@ -0,0 +1,159 @@
...
+            if (cachedDeps.find(src) == cachedDeps.end()) {
+                std::set<const SVF::VFGNode*> deps;
+                traverseMemDependencies(src, deps);
+                cachedDeps[src] = deps;
+            }
+
+            // Convert VFGNode -> llvm::Value
+            std::vector<const llvm::Value*> valDeps;
+            std::set<const SVF::VFGNode*> &memDeps = cachedDeps[src];
```
Technius, line 43: nit: llvm::SmallSet is an efficient drop-in replacement for `std::set` ([see here](https://llvm.org/docs/ProgrammersManual.html#set-like-containers-std-set-smallset-setvector-etc))

lib/analysis/interproc-analysis/SVFUtils.h
```diff
@@ -0,0 +1,159 @@
...
+            std::transform(memDeps.begin(), memDeps.end(), std::back_inserter(valDeps),
+                           [](auto mDep) { return mDep->getValue();});
+
+            // Filter out null values
+            std::vector<const llvm::Value*> nonNullValDeps;
+            std::copy_if(valDeps.begin(), valDeps.end(), std::back_inserter(nonNullValDeps),
+                         [](auto val) { return val != nullptr;});
+            return std::move(nonNullValDeps);
```
Technius, line 51: nit: might be better to use `std::remove_if` and `std::erase` on `valDeps` to avoid copying elements to `nonNullValDeps`.

lib/analysis/interproc-analysis/SVFUtils.h
```diff
@@ -0,0 +1,159 @@
...
+                auto deps = cachedDeps[src];
+                return std::any_of(deps.begin(), deps.end(), [&trg](auto node) { return node->getValue() == trg; });
+            }
+        }
+
+        static std::vector<const llvm::Value*> valueMemDeps(const llvm::Value *src) {
+            if (cachedDeps.find(src) == cachedDeps.end()) {
+                std::set<const SVF::VFGNode*> deps;
+                traverseMemDependencies(src, deps);
+                cachedDeps[src] = deps;
```
Technius, line 38: nit:
```suggestion
                cachedDeps[src] = std::move(deps);
```

lib/analysis/interproc-analysis/SVFUtils.h
Technius: What is "trg"?

lib/Vanguard.cpp
```diff
@@ -114,9 +118,40 @@ void initializeLLVM(int argc, char **argv) {
...
+    std::vector<std::string> moduleNameVec;
+    SVF::LLVMUtil::processArguments(argc, argv, arg_num, arg_value, moduleNameVec);
+
+    // Hack to avoid the const-ness of Options in SVF.
+    // We could pass additional arguments in the processArguments above, but this one is a bit cleaner.
+    *const_cast<llvm::cl::opt<bool>*>(&SVF::Options::PStat) = false;
+
+    // Pass all modules to the set. Unfortunately, it seems that SVF only supports a single set per run.
+    // I am not sure if we can have more control over this.
+    SVF::LLVMModuleSet *modSet = SVF::LLVMModuleSet::getLLVMModuleSet();
```
Technius, line 133: Note: `SVF::LLVMModuleSet::getLLVMModuleSet()` references a global variable: https://github.com/SVF-tools/SVF/blob/master/lib/SVF-FE/LLVMModule.cpp#L72

Unforuntately, `getLLVMModuleSet()` appears to be called in many places in SVF, so I don't think we can avoid this global variable (e.g., by constructing our own module set and passing it around) unless we patch SVF. And I'd rather not add more patches unless we can get some of our existing ones upstreamed...

lib/Vanguard.cpp
```diff
@@ -114,9 +118,40 @@ void initializeLLVM(int argc, char **argv) {
...
+    SVF::SVFModule* svfModule = modSet->buildSVFModule(moduleNameVec);
```
Technius, line 134: Note: it appears that `buildSVFModule()` will create its own `llvm::LLVMContext` (see [here](https://github.com/SVF-tools/SVF/blob/377b196c456cb88eb395ca1b6cf1867b75f8eb1b/lib/SVF-FE/LLVMModule.cpp#L278) -- called by `loadModules()` which is called by `buildSVFModule()`). We will have to make sure that we use this context everywhere instead of creating another one (e.g., on line 206), otherwise we may run into subtle bugs when using LLVM classes/types.

lib/analysis/interproc-analysis/SVFUtils.h
```diff
@@ -0,0 +1,159 @@
...
+
+        static SVF::BVDataPTAImpl *pa;
+
+        static SVF::PTACallGraph *callgraph;
+
+        static SVF::SVFG *svfg;
+
+        static std::map<const llvm::Value*, std::set<const SVF::VFGNode*>> cachedDeps;
+
+        static void initSVFModule() {
```
Technius, line 67: These functions should probably go in the cpp file.

lib/analysis/interproc-analysis/SVFUtils.h
kferles: You can think of this as `src->dependsOn(trg)`

lib/analysis/interproc-analysis/SVFUtils.h
kferles: Yes, I am aware. Jon complained about using the llvm data structures in an earlier commit because they force LLVM as a dependency. I prefer those as well, but let's come to an agreement and be consistent moving on.

lib/analysis/interproc-analysis/SVFUtils.h
kferles: Cool, thanks! I am not very familiar with the C++ functional interfaces.

lib/Vanguard.cpp
kferles: Yeah, I know. I think we have to stick with this one for a while. I designed it in a way so that analyses are run lazily, so if no detectors need SVF analyses there won't be much overhead.

lib/Vanguard.cpp
kferles: Good catch, I'll how I can get this out of there.

kferles:
@Technius: I addressed all the nits and left comments for whatever I did not. I also did some minor refactoring in `SFVUtils` (the last two commits).

kferles:
Ok, I think now we are good to go. Although the build check seems to be stuck.

Technius (approved):

PR #35, opened 2022-10-13
Add Yul frontend support to vanguard_driver.py

Technius:
This calls the Yul frontend from the vanguard_driver.py script, so that we can run end-to-end starting from Solidity files.

~~The Yul frontend is currently called from a "helper" script which must be manually created, since we haven't figured out a strategy for how we want to invoke this (private) dependency yet.~~ The Yul frontend is available either through Nix or by manual compilation. You can also combine a manual build of Vanguard with a version of pyul invoked by Nix if you create a wrapper script for pyul that is added to PATH.

Fixes VAN-96


Technius:
Main blocker for this is our strategy for how we want to incorporate the Yul frontend.

vanguard_driver.py
```diff
@@ -86,57 +123,95 @@ def is_cached(out_path: Path, src_path: Path) -> bool:
...
+                )
+                bc_path = sol_out_dir / (contract_name + "_deploy.bc")
+
+                if contract_name not in summaries:
+                    continue
+
+                sol_out_dir.mkdir(parents=True, exist_ok=True)
+
+                # Generate .bc file
+                subprocess.run(["llvm-as", ll_file, "-o", bc_path], check=True)
```
kferles, line 200: Is this step necessary? I think Vanguard will work fine with ll files.

kferles (approved):
I only left a minor comment, but this looks good to me.

vanguard_driver.py
Technius: Probably not, LLVM's module loading functions should be able to load ll or bc files.

PR #36, opened 2022-10-20
Fix some issues with the build

Technius:


kferles (approved):
LGTM

CMakeLists.txt
```diff
@@ -12,7 +12,14 @@ endif()
 find_package(LLVM 13 REQUIRED CONFIG)
 separate_arguments(LLVM_DEFINITIONS_LIST NATIVE_COMMAND ${LLVM_DEFINITIONS})
 add_definitions(${LLVM_DEFINITIONS_LIST})
-include_directories(${LLVM_INCLUDE_DIRS})
```
kferles, line 15: I tried rebasing on this branch and this change causes issues when there are multiple LLVM installations. The problem is that if there is an LLVM installation in PATH that is different than the LLVM_HOME one, then the compiler will use those header files which is problematic, especially with LLVM.

kferles (changes requested):

PR #41, opened 2022-12-12
Support for circom

0xddom:
Since I have already a hefty amount of code for supporting circom Im creating this pull request to start reviewing.

@Technius, the file you care about is `nix/svf/compile-flags.patch`.


aux-build-scripts/build-svf.sh
```diff
@@ -1,16 +1,22 @@
...
+function die() {
+  echo "$@"
+  exit 1
+}
+
 SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 SVF_INSTALL_DIR=${SCRIPT_DIR}/../deps/SVF/
 
 # Download SVF
+rm -rf ${SCRIPT_DIR}/SVF  # Remove whatever is there to make sure we always get a fresh copy
```
Technius, line 12:
```suggestion
rm -rf "${SCRIPT_DIR}"/SVF  # Remove whatever is there to make sure we always get a fresh copy
```

aux-build-scripts/build-svf.sh
```diff
@@ -1,16 +1,22 @@
...
 git clone https://github.com/SVF-tools/SVF.git ${SCRIPT_DIR}/SVF
-cd ${SCRIPT_DIR}/SVF
+cd ${SCRIPT_DIR}/SVF || die "Failed to cd into ${SCRIPT_DIR}/SVF"
```
Technius, line 14:
```suggestion
cd "${SCRIPT_DIR}"/SVF || die "Failed to cd into ${SCRIPT_DIR}/SVF"
```

aux-build-scripts/build-svf.sh
```diff
@@ -19,11 +25,11 @@ done
...
+cd ./'Release-build' || die "Failed to cd into ./Release-build"
 cmake -D CMAKE_INSTALL_PREFIX:PATH=${SVF_INSTALL_DIR} ../
 # TODO: Add option for job num
 make -j 8
 make install
 
-cd ${SCRIPT_DIR}
-rm -rf ${SCRIPT_DIR}/SVF
+cd ${SCRIPT_DIR} || die "Failed to cd into ${SCRIPT_DIR}"
+rm -rf ${SCRIPT_DIR}/SVF  # Cleanup
```
Technius, line 35:
```suggestion
cd "${SCRIPT_DIR}" || die "Failed to cd into ${SCRIPT_DIR}"
rm -rf "${SCRIPT_DIR}"/SVF  # Cleanup
```

lib/detectors/UnderConstraints/UnderConstraints.h
```diff
@@ -0,0 +1,90 @@
...
+
+    template<typename Domain>
+    class UnderConstraints : public UniverseDetector<Domain> {
+    public:
+        std::vector<Requirement *> registerAnalyses() override {
+            return {};
+        }
+
+        void startDetection() override {
+            analysis = new TaintAnalysis();
```
Technius: Would be better to use `std::unique_ptr` and `std::make_unique` to reduce the chance of introducing memory errors.

lib/detectors/UnderConstraints/UnderConstraints.h
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        void detect(typename Domain::Universe &universe) override {
+            for (auto unit : universe.units())
+                for (auto component : unit->components())
```
Technius, line 28: nit: explicitly indicating pointer on auto helps with readability
```suggestion
            for (auto *unit : universe.units())
                for (auto *component : unit->components())
```

lib/detectors/UnderConstraints/UnderConstraints.h
```diff
@@ -0,0 +1,90 @@
...
+                    detectInComponent(component);
+        }
+
+        void report() override {
+            std::cerr << "==== UnderConstraints analysis ====" << std::endl;
+            for (const auto& [component, signals] : underConstrainedSignals) {
+                std::cerr << "For component " << component->name() << " we have " << signals.size() << " unconstrained output signals:" << std::endl;
+                for (auto signal : signals) {
```
Technius, line 36: nit:
```suggestion
                for (auto *signal : signals) {
```

lib/detectors/UnderConstraints/UnderConstraints.h
```diff
@@ -0,0 +1,90 @@
...
+            return analysis::InterProceduralAnalyses<Domain>::valueDependsOn(out, in);
+        }
+
+        bool checkSignal(typename Domain::template Signal<llvm::Value> *out, std::vector<typename Domain::template Signal<llvm::Value> *> ins) {
+            std::cerr << "[DEBUG]  outputSignal: " << out->name() << std::endl;
+            bool ret = false;
+            for (auto in : ins) {
+                if (checkIfSignalsAreConnected(in, out)) {ret = true; goto end;}
+            }
+            end:
```
Technius, line 79: nit: can avoid the goto by using `break;`

lib/domain/libCircuit/CircuitSummaryReader.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+#include <utility>
+
+#define requireString(s, msg) require(val.HasMember(s) && val[s].IsString(), msg)
+#define requireBool(s, msg) require(val.HasMember(s) && val[s].IsBool(), msg)
+
+// Available compilers
+#define CIRCOM_COMPILER "circom"
+#define RUSTC_COMPILER "rustc"
+#define HALO2_FRAMEWORK "halo2"
```
Technius, line 18: nit: may be better to define these as `constexpr auto` so they can be better handled by IDE, type checker, etc.

lib/domain/libCircuit/CircuitSummaryReader.h
```diff
@@ -0,0 +1,68 @@
...
+            return foundComponents;
+        }
+
+        [[nodiscard]] inline const std::vector<typename CircuitDomain::UserFunction *>& functions() const {
+            return foundFunctions;
+        }
+    private:
+        const llvm::Module &module;
+        CircuitLanguageFactory * langFactory;
+        CircuitModel<CircuitDomain> *model;
```
Technius, line 43: Since these seem to be owned by the `CircuitSummaryReader`, would be better to define these as `std::unique_ptr` to avoid potential memory safety issues.

lib/detectors/UnderConstraints/UnderConstraints.h
```diff
@@ -0,0 +1,90 @@
...
+            for (auto outputSignal : component->outputSignals())
+                if (!checkSignal(outputSignal, inputSignals))
+                    addUnderConstrainedSignal(component, outputSignal);
+
+        }
+
+        void addUnderConstrainedSignal(typename Domain::Component *component, typename Domain::template Signal<llvm::Value> *out) {
+            if(underConstrainedSignals.find(component) == underConstrainedSignals.end()){
+                underConstrainedSignals[component] = std::vector<typename Domain::template Signal<llvm::Value> *>();
+            }
```
Technius, line 64: nit: I don't think this is needed; `std::map::operator[]` will initialize the entry at the given index if it does not exist yet.

lib/domain/libCircuit/langs/circom/CircomModel.h
```diff
@@ -0,0 +1,134 @@
...
+//            virtual void visit(const BinaryOpIns<Domain> &v) {};
+//            virtual void visit(const UnaryOpIns<Domain> &v) {};
+//            virtual void visit(const CallIns<Domain> &v) {};
+//            virtual void visit(const CastIns<Domain> &v) {};
+//            virtual void visit(const TernaryIns<Domain> &v) {};
+//            virtual void visit(const AssignIns<Domain> &v) {};
+            virtual void visit(const UnknownIns<Domain> &v) {
+                int probe = 0;
+                if (!llvm::isa<llvm::GetElementPtrInst>(v.unwrap())) return;
+                auto llvmInst = llvm::dyn_cast<llvm::GetElementPtrInst>(v.unwrap());
```
Technius, line 45: nit: `llvm::dyn_cast` returns null if it is not the given type, so this could be more cleanly expressed as:
```suggestion
                auto llvmInst = llvm::dyn_cast<llvm::GetElementPtrInst>(v.unwrap());
                if (!llvmInst) return;
```

lib/domain/libCircuit/langs/circom/CircomUtils.cpp
```diff
@@ -0,0 +1,35 @@
+//
+// Created by Daniel Dominguez on 30/11/2022.
+//
+
+#include "CircomUtils.h"
+#include <sstream>
+
+#define STRUCT_TEMPLATE_CIRCUIT "struct_template_circuit_"
+#define METADATA_SUFFIX "_metadata"
```
Technius, line 9: nit: can avoid macros by instead using defining as `constexpr auto`

nix/svf/compile-flags.patch
```diff
@@ -1,14 +1,21 @@
...
-
+ 
 -        set(COMMON_FLAGS "-fPIC -Werror -Wall")
 +        set(COMMON_FLAGS "-fPIC")
-
+ 
          if(CMAKE_BUILD_TYPE MATCHES "Debug")
-             set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${COMMON_FLAGS} -O0 -fno-rtti")
-
+-            set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${COMMON_FLAGS} -O0 -fno-rtti")
```
Technius, line 13: Just curious, what was the issue with RTTI?

lib/Vanguard.cpp
```diff
@@ -258,6 +259,35 @@ int main(int argc, char **argv) {
...
     }
+    else if (domain == vanguard::Detector::CIRCUIT) {
+        auto &factory = vanguard::CircuitDomain::Factory::instance();
+        std::vector<vanguard::CircuitDomain::CompilationUnit *> units;
+        units.reserve(modules.size());
+        for(auto &mod : modules) {
+            std::string filename = filenames[mod];
+
+            // summary file is filename with the extension replaced with .json
+            std::string summary = filename;
```
kferles, line 270: Lines 270 - 283 are shared between Circuits and Blockchain domains. Let's extract a function and call it in both places.

lib/analysis/interproc-analysis/SVFUtils.cpp
```diff
@@ -93,13 +93,21 @@ namespace analysis {
         SVF::FIFOWorkList<const SVF::VFGNode*> worklist;
         worklist.push(vNode);
 
+        std::cerr << "[DEBUG] Starting worklist algorithm (src=" << src << ") (trg=" << trg << ")" << std::endl;
```
kferles, line 96: Let's use a logging framework. If I am not mistaken, @Technius has used LLVM's logging API in other places. Since we are sticking with LLVM for the foreseeable future, I am okay with using that.

lib/domain/libCircuit/ComputableUnit.h
```diff
@@ -0,0 +1,31 @@
...
+
+    template<typename Base, typename Domain>
+    class Circuit<Base,Domain>::ComputableUnit : ObjectType<Domain> {
+    public:
+        explicit ComputableUnit(typename Domain::Function *logic) : logic(logic), ObjectType<Domain>() {}
+
+        const llvm::Type *unwrap() const {
+//            const llvm::Function *fn = logic->unwrap();
+//            return dynamic_cast<llvm::Type *>(fn);
+            return nullptr;
```
kferles, line 18: This feels wrong.

nix/svf/compile-flags.patch
kferles: It wouldn't let us derive from an SVF class because it couldn't find the type information.

kferles (changes requested):
Need some work, but ok overall. Let's make sure it compiles in the actions before merging.

In the future, let's aim for smaller PRs. It would make both of our lives easier.

lib/detectors/UnderConstraints/UnderConstraints.h
0xddom: I didn't knew you could do that. I will keep it in mind from now on

lib/detectors/UnderConstraints/UnderConstraints.h
0xddom: This was done for logging debug output, that function should just return there

0xddom:
CI fails because of syntax, not sure why

lib/domain/libCircuit/ComputableUnit.h
0xddom: Because it is. That was a left over from when I was figuring out the `Domain` class model. `ComputableUnit` is no longer a subclass of `ObjectType`. It does not need to be.

lib/domain/libCircuit/InputSignal.h
```diff
@@ -0,0 +1,28 @@
...
+#define VANGUARD_INPUTSIGNAL_H
+
+//#include "Signal.h"
+#include "Circuit.h"
+
+namespace vanguard {
+
+    template<class Base, class Domain>
+    template<typename Wrap, typename D>
+    class Circuit<Base, Domain>::InputSignal : public Circuit<Base, Domain>::template Signal<Wrap, D> {
```
kferles, line 15: The issue with the build might be that InputSignal is inheriting from a template class. I am not sure why we need that in the first place.

0xddom: What I dont get is why it builds with clang and not with gcc

0xddom: Also the whole Type and Value modules are based around inheritance of template classes, I dont see why is that an issue?

kferles: We had similar issues in the past, where clang is more permissive around templates. @stephensj2: do you have any suggestions on how to resolve this?

lib/analysis/interproc-analysis/SVFUtils.cpp
Technius: I don't think there is a big problem with adding third-party libraries to this project, because Vanguard is licensed under GPLv3. Spdlog is good, but we could probably get by without it for now.

lib/domain/libCircuit/InputSignal.h
0xddom: From what I could gather from the internet the issue could be a circular dependecy. But I havent found it yet

lib/domain/libCircuit/CircuitSummaryReader.h
0xddom: This two pointers are passed around to the visitor classes and finally returned as part of the summary. I think a shared pointer would be better in this case

Technius: If I understand correctly, `langFactory` and `model` are allocated by `readSummary`, but they do not appear to be used elsewhere.

lib/domain/libCircuit/CircuitSummaryReader.cpp
```diff
@@ -0,0 +1,52 @@
...
+    void CircuitSummaryReader::readSummary(rapidjson::Value &val) {
+        requireString("compiler", "Summary must declare compiler");
+        requireString("version", "Summary must declare version");
+
+        std::string compiler = val["compiler"].GetString();
+        std::string version = val["version"].GetString();
+
+        if (compiler == CIRCOM_COMPILER) {
+            model = new CircomModel<CircuitDomain>(module, CircuitDomain::Factory::instance());
+            langFactory = new CircomFactory(CircuitDomain::Factory::instance());
```
Technius, line 40: Won't this leak memory if `readSummary` is called twice?

lib/Vanguard.cpp
```diff
@@ -141,6 +142,26 @@ LLVMContext* initializeSVF(int argc, char **argv, std::unordered_map<llvm::Modul
...
+        summary.replace(dotIdx, std::string::npos, ".json");
+    }
+
+    if (!llvm::sys::fs::exists(summary)) {
+        llvm::WithColor::error()
+                << "the summary file " << summary << " does not exist!\n";
+        throw std::runtime_error("Summary file does not exist");
+    }
+
+    llvm::dbgs() << "Read summary from: " << summary << "\n";
```
Technius, line 159: Quick note: ideally we would `#define DEBUG_TYPE "vanguard"` at the top of the file and wrap the debug code in `LLVM_DEBUG({ llvm::dbgs() << "foo\n"; });` blocks. This will allow us to add an LLVM opt flag like `-debug` to enable debug output; however this functionality requires us to either 1) build LLVM with assertions & ABI breaking checks enabled; or 2) create a custom debug flag and set `llvm::DebugFlag = true` (see `llvm/Support/Debug.h`).

lib/domain/libCircuit/CircuitSummaryReader.cpp
0xddom: Technically yes, but `readSummary` is private and is called from the constructor. So while not enforced there is only one instance of each pointer per `CircuitSummaryReader` instance

lib/domain/libCircuit/langs/circom/CircomModel.h
```diff
@@ -29,6 +29,7 @@ namespace vanguard {
         }
 
         const llvm::Value *getSignalValue(const typename Domain::template Signal<llvm::Value> *signal) override {
+            // TODO: This algorithm should work only on Input Signals
 #define setAndFailIfNull(v, e) v = e; if (v == nullptr) return nullptr
```
kferles, line 34: I would suggest getting rid of macros. If this is an assumption for correctness, then let's add assertions here.

lib/domain/libCircuit/langs/circom/CircomSearchers.h
```diff
@@ -0,0 +1,141 @@
...
+            result = newResult;
+            done = true;
+        }
+    private:
+        std::optional<Result> result;
+        bool done;
+    };
+
+    template <typename Domain, typename Result>
+    std::optional<Result> searchOne(typename Domain::Function *func, InstructionClassQuery<Domain, Result> &searchQuery) {
```
kferles, line 36: Is the intention here to stop after we find the first match? If so, let's rename the function to "findFirst". Similar to "findAll" for "searchMany".

lib/domain/libCircuit/langs/circom/CircomModel.h
```diff
@@ -65,6 +65,7 @@ namespace vanguard {
 
         std::vector<typename Domain::template Constraint<llvm::Value> *> getComponentConstraints(typename Domain::Function *componentLogic) override {
             std::vector<typename Domain::template Constraint<llvm::Value> *> constraints;
+#if 0
```
kferles, line 68: Should this preprocessor branch be removed?

kferles (changes requested):
Once we address these and fix the output signal stuff, I believe we can merge and start prototyping detectors.

lib/domain/libCircuit/langs/circom/CircomSearchers.h
0xddom: Could be. The responsibility of what is returned falls onto the query object. From the function's perspective it will return either the first value returned by the query or all the query could find, so I can see either set of names.

Also, I think this two functions could be part of the `Domain::Function` class and the specialised visitor added to `Instruction.h` with its parent class. This way others could benefit from the API as it is general enough.

0xddom:
@kferles All done from my side

kferles (approved):

kferles:
Why did we merge v1 circom-llvm-testing? Shouldn't it go the other way around? I am not sure what would be the consequences of this. Other than that, it's fine to merge by me. @stephensj2: are you ok with merging into v1?

0xddom:
Yeah, there was a merge conflict in Vanguard.cpp that I had to address

PR #43, opened 2022-12-29
Nicholas/divide before multiply

Sapo-Dorado:
Creates a new detector, `DivideBeforeMultiplyDetector`, that checks whether a function applies a multiplication operation on the result of a division operation. Also fixes a typo in `vanguard_driver.py` for the reentrancy detector.


lib/detectors/Statements/DivideBeforeMultiplyDetector.h
```diff
@@ -0,0 +1,120 @@
...
+                    }
+                }
+            }
+        }
+
+        void detect(typename Domain::Universe &universe) override {
+            auto contracts = universe.contracts();
+            for (auto contract : contracts) {
+                for (auto fn : contract->functions()) {
+                    MulDivResolver* resolver = new MulDivResolver();
```
Technius, line 90: nit: use `std::unique_ptr`/ `std::make_unique` to reduce the chance of memory errors and avoid manual `delete`

lib/detectors/Statements/DivideBeforeMultiplyDetector.h
```diff
@@ -0,0 +1,120 @@
...
+
+
+        void process(Function *fn, MulDivResolver *mulDivResolver) {
+            for (auto blk: fn->blocks()) {
+                for (auto ins: blk->instructions()) {
+                    ins->accept(*mulDivResolver);
+
+                    CallTargetResolver callResolver;
+                    ins->accept(callResolver);
+                    for(auto tgt : callResolver.tgts) {
```
Technius, line 75: If the function is the selector, then the targets will be every public function in the module. So in effect, the detector is flagging at the module level rather than at the function level.

I think this method should skip the function if it is not a public function.

Technius:
The build failure seems to be some bug in pyul. @veriAjinkya could you take a look?

Sapo-Dorado:
Summary of running tests:
[https://www.notion.so/veridise/DivideBeforeMultiply-Testing-Results-d56dc453703f410abfccf48e35512fd2](https://www.notion.so/veridise/DivideBeforeMultiply-Testing-Results-d56dc453703f410abfccf48e35512fd2)

Technius:
Last missing piece is automated testing through lit. Please see `test/hello.txt` for an example.

Sapo-Dorado:
> Last missing piece is automated testing through lit. Please see `test/hello.txt` for an example.

added

test/DivideBeforeMultiply.txt
```diff
@@ -0,0 +1 @@
+// RUN: python3 vanguard_driver.py --detector DivideBeforeMultiply -D %S/Inputs --src_path %S/Inputs/DivideBeforeMultiply.sol
```
Technius, line 1: Actually, I would move this comment into the source file, and then move the source file to the top level `test/` folder so that lit can discover the test automatically. Then the command can be changed to:
```sh
python3 vanguard_driver.py --detector DivideBeforeMultiply --src_path %s | FileCheck %s
```
This is better than just using lit directly (which only checks the exit code), because [FileCheck](https://llvm.org/docs/CommandGuide/FileCheck.html) can be used to match regexes against the output.

Sapo-Dorado: > Actually, I would move this comment into the source file, and then move the source file to the top level `test/` folder so that lit can discover the test automatically. Then the command can be changed to:
> 
> ```shell
> python3 vanguard_driver.py --detector DivideBeforeMultiply --src_path %s | FileCheck %s
> ```
> 
> This is better than just using lit directly (which only checks the exit code), because [FileCheck](https://llvm.org/docs/CommandGuide/FileCheck.html) can be used to match regexes against the output.

Ok, updated the test and used FileCheck to ensure that the output is what we expect

test/DivideBeforeMultiply.sol
```diff
@@ -0,0 +1,63 @@
+// RUN: python3 vanguard_driver.py --detector DivideBeforeMultiply -D %S --src_path %s | FileCheck %s
+// CHECK: DivideBeforeMultiplyVulnerable: vulnerable
+// CHECK: Check the following functions for vulnerabilities:
+// CHECK: {{vulnerable[123]}}
+// CHECK: {{vulnerable[123]}}
+// CHECK: {{vulnerable[123]}}
```
Technius, line 6: Currently, this only checks that vulnerable functions are reported, and does not check that the not-vulnerable functions are unreported. I would also add additional lines `CHECK-NOT: notVulnerable` [1] for the not-vulnerable functions. 

[1] between the checks (see the section on "CHECK-NOT" in the FileCheck help) You can pass the `--implicit-check-not <pattern>` flag to avoid having to insert a bunch of CHECK-NOT.

Technius (approved):
I think this should be ready to squash and merge after addressing this one last nitpick I have.

Sapo-Dorado:
> I think this should be ready to squash and merge after addressing this one last nitpick I have.

Ok, I added an --implicit-check-not that looks for the pattern `notVulnerable`. This should catch any of the functions with notVulnerable in the name being reported as vulnerable.

PR #44, opened 2023-01-03
Under constrained output detector [VAN-114]

0xddom:


0xddom:
@kferles nobug.circom and bug.circom pass but condbug.circom does not.

I'm thinking that maybe we should encode this as unit tests and start prototyping an architecture that make detectors testeable. I think we just need to create a separate layer for presentation (i.e printing to stderr) that we can mock in unit tests. Questions is, a different PR or in this PR?

0xddom:
We seem to hit a global even with the change we did yesterday. And I'm not sure why

0xddom:
Only thing missing are e2e tests with `lit`. Code is ready.

aux-build-scripts/build-pyul.sh
```diff
@@ -25,6 +25,7 @@ git clone https://github.com/Veridise/yul2llvm.git "$BUILD_DIR"
 
 # Build Python part
 cd "$BUILD_DIR"
+git checkout bryan/fix-llvm-dep
```
kferles, line 28: I think this has been merged into main now.

lib/analysis/taint-analysis/TaintAnalysis.h
```diff
@@ -44,6 +44,76 @@ namespace vanguard {
...
+
+        SrcsToSnks::iterator end() {
+            return srcsToSnks.end();
+        }
+
+        std::vector<const llvm::Value *> operator[](const llvm::Value *val) {
+            return srcsToSnks[val];
+        }
+
+        void onUnreachable(SVF::ProgSlice *slice) override {
```
kferles, line 66: Why do both reachable and unreachable pairs are stored on the same map?

lib/domain/libCircuit/langs/circom/CircomModel.h
```diff
@@ -139,26 +139,26 @@ namespace vanguard {
...
-        }
-
-        llvm::NamedMDNode *searchStructMetadata(std::string componentName) {
-            std::string metadataName = componentNameToStructMetadataName(std::move(componentName));
-
-            auto metadata = module.getNamedMetadata(metadataName);
-            if (!metadata) return nullptr;
-            return metadata;
-        }
+//        int signalNameToIndex(const std::string& signalName, llvm::NamedMDNode *metadata) {
```
kferles, line 142: If we no longer need this, let's delete it.

lib/domain/libCircuit/langs/circom/CircomSearchers.h
```diff
@@ -116,10 +118,45 @@ namespace vanguard {
...
     private:
         typename Domain::Function *target;
     };
 
+    template <typename Domain>
+    class SearchConstraintStoreOperation : public InstructionClassQuery<Domain, const llvm::Value *> {
+    public:
+        explicit SearchConstraintStoreOperation(const llvm::Value *constraintMem) : mem(constraintMem) {};
+
+//        void visit(const UnknownIns<Domain> &v) override {
```
kferles, line 134: Same here! Delete if not needed.

vanguard_driver.py
```diff
@@ -276,6 +322,31 @@ def run_detector(detector: str, summary_path: Path, out_dir: Path) -> Dict[str,
     return results
 
 
+def run_detector_on_ll(detector: str, summary: Path, out_dir: Path) -> Dict[str, str]:
```
kferles, line 325: why do we need separate functions for bc and ll?

kferles (changes requested):
Looks good in general. Just a few minor things.

vanguard_driver.py
0xddom: Because of the glob. It searches `*_deploy.bc` files instead of `*.ll`. Maybe it can be implemented with a postfix argument for the extension instead than 2 separate functions.

lib/analysis/taint-analysis/TaintAnalysis.h
0xddom: Because this observer just collects all the slices. An unreachable slice will not have sinks, which is the part I care about in the detector.

aux-build-scripts/build-pyul.sh
0xddom: Yes, I can remove that now

kferles (approved):
LGTM

PR #46, opened 2023-01-11
Basic SARIF support in Vanguard [VAN-128] [VAN-135]

0xddom:
Draft of the SARIF module. I'm creating this PR early so we can discuss the design.


lib/detectors/SARIF/SARIFLocation.h
```diff
@@ -0,0 +1,39 @@
...
+
+#ifndef VANGUARD_SARIFLOCATION_H
+#define VANGUARD_SARIFLOCATION_H
+
+#include "SARIFObject.h"
+#include <optional>
+#include <vector>
+#include <string>
+
+namespace vanguard {
```
kferles, line 13: Can we add some comments above each class that describes the differences between these type of locations?

lib/detectors/SARIF/SARIFLog.h
```diff
@@ -0,0 +1,21 @@
...
+#include "SARIFObject.h"
+#include "SARIFRun.h"
+#include "vector"
+
+namespace vanguard {
+    class SARIFLog : public SARIFObject {
+    public:
+        void addRun(std::unique_ptr<SARIFRun> run);
+    private:
+        std::vector<std::unique_ptr<SARIFRun>> runs;
```
kferles, line 17: Out of curiosity: what does a SARIFRun represent?

lib/detectors/SARIF/SARIFLocation.h
0xddom: Yes! The standard has a good explanation of how they work. We can put it in the code for reference

lib/detectors/SARIF/SARIFLog.h
0xddom: The execution of a particular tool. This way you can merge results of different tools into a single file

0xddom:
Broadly, my design goals are:
- Have an object model that incrementally replicates SARIF but in a more readable manner
- Have the bare minimum we need to be considered as SARIF compliant in a first iteration and then expand on more SARIF features as we need
- Separate the object model from the serialization to potentially support multiple formats
  - Imagine a future where we decide to write the reports in BSON from the tools to the backend and reencode into JSON to the frontend
- Encapsulate the reporting inside the base `Detector`
  - Offer a public API to the subclasses (i.e. `Detector::alert(string message, string file, int line, int col, string name)`)
  - Detectors  would not need to override `report` as that method would change to just report all the results found.

kferles:
This looks good to me on a high level. One note: I would suggest having a DetectorAlert abstract class with a virtual method called toSARIFResult. This will give detectors the flexibility to create custom alerts and implement their conversion to SARIF. So, the public API you propose could be Detector::alert(DetectorAlert alert).

Technius:
I think a better location for the code is in `lib/SARIF`, and not `lib/detectors/SARIF`.

Technius:
It would be very helpful if you could set up a simple unit test in, say, `unit-tests/SARIFTests.cpp` to mock up how the API might be used.

0xddom:
Yeah, missing headers in gcc that clang does not complain about

0xddom:
So far you can tell how the serialization is intended to work

lib/SARIF/SARIF.h
```diff
@@ -0,0 +1,26 @@
...
+// Created by Daniel Dominguez on 11/1/23.
+//
+
+#ifndef VANGUARD_SARIF_H
+#define VANGUARD_SARIF_H
+
+namespace vanguard {
+
+    template<typename OutFmt>
+    class SARIF {
```
Technius, line 11: This seems overly complicated. The data structure that represents the report/document should be independent of the file format(s) that it can be exported to.

Technius (changes requested):

Technius:
By the way, there seems to be an [official SARIF library for Python](https://github.com/microsoft/sarif-python-om), so we only need to implement the SARIF data structures for the C++ part of Vanguard.

0xddom:
> By the way, there seems to be an [official SARIF library for Python](https://github.com/microsoft/sarif-python-om), so we only need to implement the SARIF data structures for the C++ part of Vanguard.

Yes, I checked for TPLs first. I plan to use it in the driver. Sadly no C/C++ version..

0xddom:
All ready from my side

lib/Vanguard.cpp
```diff
@@ -158,7 +184,7 @@ std::unique_ptr<SummaryReader> loadSummary(std::vector<llvm::Module *> modules,
     }
 
     llvm::dbgs() << "Read summary from: " << summary << "\n";
-    const llvm::Module *module = modules.back();
+//    const llvm::Module *module = modules.back();
```
kferles, line 187: Let's delete this comment.

vanguard_driver.py
```diff
@@ -336,12 +358,13 @@ def run_detector(detector: str, summary_path: Path, out_dir: Path, postfix: str)
 #    print(tabulate(out, tablefmt="grid"))
 
 
-def print_results(results):
-    for cls, ress in results.items():
-        for info in ress:
-            detector, res = info
-            print("--------- {} ---------".format(detector))
-            print(res)
+# def print_results(results):
```
kferles, line 361: Delete as well.

vanguard_driver.py
```diff
@@ -427,15 +452,30 @@ def main():
...
     args = parser.parse_args()
 
     if not args.src_path.exists():
         sys.exit(f"Source path {args.src_path} does not exist!")
 
-    if args.detector not in DETECTORS and args.detector != "all":
-        sys.exit(
-            f"{args.detector} is not a recognized detector (check DETECTORS at top of run.py)!"
-        )
+    # if args.detector not in DETECTORS and args.detector != "all":
```
kferles, line 475: Delete as well.

kferles (changes requested):
Just delete the commented code and we are good to go!

unit-tests/CMakeLists.txt
```diff
@@ -11,14 +11,15 @@ macro(vanguard_gtest_suite target)
 endmacro()
 
 add_executable(VanguardUnitTests
-    vanguard-unittests.cpp
+    vanguard-unittests.cpp sarif-unittests.cpp
 )
 
 target_include_directories(VanguardUnitTests PRIVATE
-  "${CMAKE_CURRENT_BINARY_DIR}"
+  "${CMAKE_CURRENT_BINARY_DIR}" "${CMAKE_INSTALL_INCLUDEDIR}"
```
Technius, line 18: Why is `CMAKE_INSTALL_INCLUDEDIR` needed here? Seems like `CMAKE_CURRENT_BINARY_DIR` should be sufficient.

unit-tests/sarif-unittests.cpp
```diff
@@ -0,0 +1,153 @@
+//
+// Created by Daniel Dominguez on 12/1/23.
+//
+
+#include <gtest/gtest.h>
+#include "unittests_config.h"
+
+#include "../lib/SARIF/SARIF.h"
+#include "../lib/SARIF/Message.h"
```
Technius, line 9:
```suggestion
#include "SARIF/SARIF.h"
#include "SARIF/Message.h"
```

lib/SARIF/SARIF.h
```diff
@@ -0,0 +1,31 @@
+//
+// Created by Daniel Dominguez on 11/1/23.
+//
+
+#ifndef VANGUARD_SARIF_H
+#define VANGUARD_SARIF_H
+
+namespace vanguard {
+
+    class SARIF {
```
Technius, line 10: nit: I think this can be replaced with `namespace SARIF` since there are no fields, and there is no code that creates an instance of a SARIF class.

lib/SARIF/formats/JSONSerializer.cpp
```diff
@@ -0,0 +1,182 @@
...
+                delete runJson;
+            }
+            o->AddMember("runs", *runs, document->GetAllocator());
+            delete runs;
+        }
+        return o;
+    }
+
+    inline rapidjson::Value *JSONSerializer::emptyObject() {
+        return new rapidjson::Value(rapidjson::kObjectType);
```
Technius, line 170: nit: why not return `std::unique_ptr` for these methods?

lib/SARIF/Log.h
```diff
@@ -0,0 +1,36 @@
...
+        using Run = SARIF::Run;
+        void addRun(std::unique_ptr<Run> run) {
+            runs.push_back(std::move(run));
+        }
+
+        std::string getVersion() {
+            return SARIF_VERSION;
+        }
+
+        std::vector<std::unique_ptr<Run>> getRuns() {
```
Technius, line 27: nit: rename to indicate that the variable is being moved.
```suggestion
        std::vector<std::unique_ptr<Run>> takeRuns() {
```

lib/SARIF/Log.h
```diff
@@ -0,0 +1,36 @@
...
+
+namespace vanguard {
+    class SARIF::Log : public SARIF::Object {
+    public:
+        using Run = SARIF::Run;
+        void addRun(std::unique_ptr<Run> run) {
+            runs.push_back(std::move(run));
+        }
+
+        std::string getVersion() {
```
Technius, line 23: nit: 
```suggestion
        std::string getVersion() const {
```

lib/SARIF/Message.h
```diff
@@ -0,0 +1,26 @@
...
+#include "Serializer.h"
+#include <string>
+#include <utility>
+
+namespace vanguard {
+
+    class SARIF::Message : public SARIF::Object {
+    public:
+        explicit Message(std::string message) : message(std::move(message)) {};
+        std::string getText() { return message; }
```
Technius, line 19: nit:
```suggestion
        std::string getText() const { return message; }
```

lib/SARIF/Result.h
```diff
@@ -0,0 +1,58 @@
...
+        Result(std::string detectorName, SARIF::Level severityLevel, std::unique_ptr<Message> message,
+               std::unique_ptr<Location> location) :
+                detectorName(std::move(detectorName)), severityLevel(severityLevel), message(std::move(message)) {
+            locations.push_back(std::move(location));
+        };
+        Result(SARIF::Result &other) :
+                detectorName(other.detectorName),
+                severityLevel(other.severityLevel),
+                message(std::move(other.message)),
+                locations(std::move(other.locations)) {};
```
Technius, line 36: This looks like a copy constructor, but it shouldn't be implemented like this because it moves the arguments. Why not leave this as the default copy constructor?

lib/SARIF/Result.h
```diff
@@ -0,0 +1,58 @@
...
+
+        std::string getDetectorName() { return detectorName; }
+        std::string getSeverityLevel() {
```
Technius, line 39: nit:
```suggestion
        std::string getDetectorName() const { return detectorName; }
        std::string getSeverityLevel() const {
```

lib/SARIF/Result.h
```diff
@@ -0,0 +1,58 @@
...
+            switch(severityLevel) {
+                case ERROR: return "error";
+                case WARNING: return "warning";
+                case NOTE: return "note";
+                case NONE: return "none";
+            }
+        }
+        std::unique_ptr<Message> getMessage() { return std::move(message); }
+        std::vector<std::unique_ptr<Location>> getLocations() { return std::move(locations); }
```
Technius, line 48: I don't think these should move the variables, otherwise there's no way to "read" the result without modifying it.
```suggestion
        const Message *getMessage() const { return message.get(); }
        const std::vector<std::unique_ptr<Location>> &getLocations() const { return locations; }
```

If we really need some sort of move method, it would be better to rename (or copy) the current implementation as `takeMessage/Locations()`.

lib/SARIF/Run.h
```diff
@@ -0,0 +1,41 @@
...
+constexpr auto SARIF_DRIVER_NAME = "Vanguard";
+
+namespace vanguard {
+    class SARIF::Run : public SARIF::Object {
+    public:
+        using Result = SARIF::Result;
+
+        void addResults(std::vector<std::unique_ptr<Result>> &newResults)  {
+            for (auto &x : newResults)
+                results.push_back(std::move(x));
```
Technius, line 23: Note that this will leave a bunch of nullptrs in `newResults`. Should also call `newResults.clear()`.

lib/SARIF/Serializer.h
```diff
@@ -0,0 +1,27 @@
...
+    template<typename OutFmt>
+    class SARIF::Serializer {
+    public:
+        virtual OutFmt serialize(SARIF::Message &message) = 0;
+        virtual OutFmt serialize(SARIF::PhysLocation &physLocation) = 0;
+        virtual OutFmt serialize(SARIF::LogicLocation &logicLocation) = 0;
+        virtual OutFmt serialize(SARIF::Location &location) = 0;
+        virtual OutFmt serialize(SARIF::Result &result) = 0;
+        virtual OutFmt serialize(SARIF::Run &run) = 0;
+        virtual OutFmt serialize(SARIF::Log &log) = 0;
```
Technius, line 22: I would make the parameters `const` because a reasonable expectation is that serialization will not modify the input data structures.
```suggestion
        virtual OutFmt serialize(const SARIF::Message &message) = 0;
        virtual OutFmt serialize(const SARIF::PhysLocation &physLocation) = 0;
        virtual OutFmt serialize(const SARIF::LogicLocation &logicLocation) = 0;
        virtual OutFmt serialize(const SARIF::Location &location) = 0;
        virtual OutFmt serialize(const SARIF::Result &result) = 0;
        virtual OutFmt serialize(const SARIF::Run &run) = 0;
        virtual OutFmt serialize(const SARIF::Log &log) = 0;
```

lib/Vanguard.cpp
```diff
@@ -63,13 +71,31 @@ void runDetectors(llvm::ModuleAnalysisManager &mam, llvm::FunctionAnalysisManage
...
     }
+    log.addRun(std::move(run));
+
+    if (!sarifReport.empty()) {
+        std::cout << "Writing SARIF report to " << sarifReport << std::endl;
+        vanguard::JSONSerializer serializer;
+        vanguard::JSONSerializer::JSONValue json = serializer.serialize(log);
+        char writeBuffer[65536];
+        FILE* fp = fopen(sarifReport.c_str(), "wb");
+        rapidjson::FileWriteStream os(fp, writeBuffer, sizeof(writeBuffer));
```
Technius, line 93: I think it should be possible to use `rapidjson::BasicOStreamWrapper<std::ofstream>` here instead to avoid the C `FILE*`-related manual resource cleanup.

lib/detectors/DetectorResult.h
```diff
@@ -0,0 +1,31 @@
...
+#include <memory>
+#include <utility>
+
+namespace vanguard {
+    class DetectorResult {
+    public:
+        explicit DetectorResult(std::string detectorName) : detectorName(std::move(detectorName)) {};
+        virtual ~DetectorResult() = default;
+        virtual std::unique_ptr<SARIF::Result> toSARIFResult() = 0;
+        std::string getDetectorName() {
```
Technius, line 21: nit:
```suggestion
        std::string getDetectorName() const {
```

Technius (commented):
Also, what is `test/sarif/sarif-2.1.0-rtm.4.json` used for?

0xddom:
> Also, what is `test/sarif/sarif-2.1.0-rtm.4.json` used for?

I think it would be of interest to add an integration test that takes reports generated by the tool and check them against the schema. I left that there to build the test and I must have get distracted with other things and forget to implement it. We need an extra tool to do this checking and it must support JSON Schema Draft-04. I havent found one that works yet.

lib/SARIF/Location.h
```diff
@@ -0,0 +1,79 @@
...
+        Location(std::unique_ptr<PhysLocation> physicalLocation, std::unique_ptr<LogicLocation> logicalLocation):
+                physicalLocation(std::make_optional(std::move(physicalLocation))) {
+            logicalLocations.push_back(std::move(logicalLocation));
+        };
+
+        [[nodiscard]] bool hasPhysicalLocation() const {
+            return physicalLocation.has_value();
+        }
+
+        [[nodiscard]] std::optional<const PhysLocation*> getPhysicalLocation() const {
```
Technius, line 63: nit: I think it's cleaner to return `PhysLocation*` (nullptr if nullopt). The additional optional seems unnecessary.

lib/SARIF/Message.h
```diff
@@ -0,0 +1,26 @@
...
+#include "Serializer.h"
+#include <string>
+#include <utility>
+
+namespace vanguard {
+
+    class SARIF::Message : public SARIF::Object {
+    public:
+        explicit Message(std::string message) : message(std::move(message)) {};
+        std::string getText() const { return message; }
```
Technius, line 19: nit:
```suggestion
        [[nodiscard]] std::string getText() const { return message; }
```

lib/SARIF/Location.h
```diff
@@ -0,0 +1,79 @@
...
+#include <string>
+#include <memory>
+
+namespace vanguard {
+    /// Represents a physical location in an artifact (i.e. A file)
+    /// It's not mandatory
+    class SARIF::PhysLocation : public SARIF::Object {
+    public:
+        PhysLocation(std::string artifactURI, int line, int column) : artifactURI(std::move(artifactURI)), line(line), column(column) {};
+        std::string getArtifactURI() const { return artifactURI; }
```
Technius, line 22: nit:
```suggestion
        [[nodiscard]] std::string getArtifactURI() const { return artifactURI; }
```

lib/SARIF/Location.h
```diff
@@ -0,0 +1,79 @@
...
+        std::string artifactURI;
+        int line, column;
+    };
+
+    /// Represents a logical location such as a XPath, a class name, a variable name, etc.
+    /// Since an object can be logically referenced with many names we can have multiples of this.
+    class SARIF::LogicLocation : public SARIF::Object {
+    public:
+        explicit LogicLocation(std::string name) : name(std::move(name)) {};
+        std::string getName() const { return name; }
```
Technius, line 35: nit:
```suggestion
        [[nodiscard]] std::string getName() const { return name; }
```

lib/SARIF/Serializer.h
```diff
@@ -0,0 +1,27 @@
...
+#define VANGUARD_SARIFSERIALIZER_H
+
+#include "SARIF.h"
+
+
+namespace vanguard {
+
+    template<typename OutFmt>
+    class SARIF::Serializer {
+    public:
```
Technius, line 15: Unless I'm mistaken, isn't a virtual destructor required?

Technius (approved):
LGTM.

kferles (approved):

Technius:
Regarding the JSON schema stuff, we can automatically generate Python classes & serialization/deserialization code using a library like https://github.com/koxudaxi/datamodel-code-generator . This also gives us validation for free.

0xddom:
Right now the Python driver does not touch sarif. Everything is handled by the c++ tool. For now we could use something like that to write a simple testing script that takes the sarif report and either passes or fails a lit test (a la FileCheck).

PR #48, opened 2023-01-14
Add Docker image builder for V1

Technius:
This adds a Docker image build by building it through Nix and then pushing to DockerHub.


Technius:
Test with:
```sh
docker run --rm -it -v "$PWD"/test:/test veridise/vanguard:git-latest -D /test --src_path /test/Inputs/HelloWorld.sol --detector statGen
```

kferles:
I am not equipped to review this PR. I tried the command above and I got the following error:

Unable to find image 'veridise/vanguard:git-latest' locally
docker: Error response from daemon: pull access denied for veridise/vanguard, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.
See 'docker run --help'.

I don't understand the changes to a level that I can be helpful.

PR #49, opened 2023-01-17
Detector for unused output signal [VAN-140]

0xddom:
Basic structure so far


Technius:
I think this needs to be rebased to be compatible with the new SARIF reporter.

0xddom:
I don't think it is mandatory. I wrote the detector part in such a way that it's backwards compatible with older detectors. They just never generate a SARIF report but the code should still compile. Otherwise I would have had to update every single detector we have so far.

stephensj2 (approved):

PR #50, opened 2023-01-17
Nicholas/unchecked return [VAN-133]

Sapo-Dorado:
create UncheckedReturnDetector


test/UncheckedReturn.sol
```diff
@@ -0,0 +1,58 @@
+// RUN: python3 vanguard_driver.py --detector UncheckedReturn -D %S --src_path %s | FileCheck %s --implicit-check-not notVulnerable
+// CHECK: unchecked_return1: vulnerable
+// CHECK: unchecked_return2: vulnerable
+// CHECK: unchecked_return3: vulnerable
+
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.10;
+
+import "./Inputs/SafeMath.sol";
```
Technius, line 9: note: this might be better to supply as an "include path" (which is not supported yet by the driver). `pyul` already supports an include path argument so we could add this easily to the driver script. Don't need to do this in this PR though.

Technius (commented):
Note: as discussed offline, the SafeMath functions are incorrectly identified as helper functions are and thus causing a failure. This seems to be caused by an issue in the driver script related to the summaries.

test/UncheckedReturn.sol
Technius: This can be included directly from `@openzeppelin`. The lit test will need to pass `--include %node_deps` to vanguard_driver.

PR #51, opened 2023-01-18
Fix Underconstrained detector according to new test cases [VAN-141]

0xddom:
This test cases are false positives due to limitations in the implementation. Need to fix the detector and/or the backend to support this cases and not report them.


0xddom:
Closing this since we are updating the frontend and tests may change

PR #54, opened 2023-01-20
Create Flashloan Detector

Sapo-Dorado:
Creates a detector that detects when the result of a `balanceOf` operation flows into the amount field of a `transfer` or a `transferFrom` operation


lib/detectors/FlashLoan/FlashloanDetector.h
```diff
@@ -0,0 +1,126 @@
...
+                            potentiallyVulnerable.insert(fnName);
+                        }
+                    }
+                }
+                vulnerable = true;
+            }
+            bool isVulnerable() {
+                return vulnerable;
+            }
+            std::unordered_set<std::string> getPotentiallyVulnerable() {
```
Technius, line 61: nit: to allow the caller to take by reference, I would change this to:
```suggestion
            const std::unordered_set<std::string> &getPotentiallyVulnerable() const {
```

test/Flashloan.sol
```diff
@@ -0,0 +1,110 @@
+// RUN: python3 vanguard_driver.py --detector Flashloan -D %S --src_path %s | FileCheck %s --implicit-check-not notVulnerable
+// CHECK: FlashloanVulnerable: vulnerable
+// CHECK: Check the following functions for vulnerabilities:
+// CHECK: {{vulnerable[123]}}
+// CHECK: {{vulnerable[123]}}
```
Technius, line 5: Can use `CHECK-DAG` to check out-of-order:
```suggestion
// CHECK-DAG: vulnerable1
// CHECK-DAG: vulnerable2
// CHECK-DAG: vulnerable3
```

lib/detectors/FlashLoan/FlashloanDetector.h
```diff
@@ -72,6 +84,12 @@ namespace vanguard {
             for (auto blk: fn->blocks()) {
                 for (auto ins: blk->instructions()) {
                     ins->accept(*resolver);
+                    CallTargetResolver callResolver;
+
+                    ins->accept(callResolver);
+                    for(auto tgt : callResolver.tgts) {
```
Technius, line 90: nit: can improve maintainability by annotating auto with `*` for pointer types
```suggestion
                    for(auto *tgt : callResolver.tgts) {
```

lib/detectors/FlashLoan/FlashloanDetector.h
```diff
@@ -19,22 +19,34 @@ namespace vanguard {
...
+                if(isTransferFunction(v.name())) {
+                    if(v.numOperands() > 0) {
+                        transferAmts.push_back(v.operandAt(v.numOperands()-2)->unwrap());
                     }
-                    transferAmts.push_back(lastArg);
                 }
             }
 
+            bool isBalanceFunction(std::string_view name) {
+                return name.find("fun_balanceOf_") != std::string::npos;
```
Technius, line 42: note: I find it weird that the name is specialized to what is generated by Yul/pyul. We'll have to come to agreement on what `CallIns<...>::name()` is supposed to return.

lib/detectors/FlashLoan/FlashloanDetector.h
```diff
@@ -19,22 +19,34 @@ namespace vanguard {
...
                 }
-                if(v.name().find("fun_transferFrom_") != std::string::npos ||
-                    v.name().find("fun_transfer_") != std::string::npos) {
-                    const llvm::Value *lastArg;
-                    for (int i = 0; i < v.numOperands() - 1 ; i++) {
-                        lastArg = v.operandAt(i)->unwrap();
-
+                if(isTransferFunction(v.name())) {
+                    if(v.numOperands() > 0) {
+                        transferAmts.push_back(v.operandAt(v.numOperands()-2)->unwrap());
```
Technius, line 36: pyul will likely generate a llvm function call of the form `pyul_call_$selector(self, ctx, arg0, arg1, ...)`.

The transfer/transferFrom cases are:
* `token.transfer(to, amount)`
* `token.transferFrom(from, to, amount)`

Shouldn't the index of `amount` be `v.numOperands() - 1`?

note: I also find it weird that the args function includes self and ctx, which are variables we do not care about.

Sapo-Dorado: In the llvm code the amount field for the `transfer` function is the fourth argument to the call instruction. From within Vanguard, calling `v.numOperands()` returns `5` for the transfer instruction. I'm trying to access the fourth argument of the function which would be `v.numOperands()-2`. I'm not sure why `v.numOperands()` is returning one higher than the number of arguments in the llvm, but that's why I've been doing -2 instead of -1.

Technius (approved):
LGTM. The operand thing is weird to me, but the tests are passing.

PR #56, opened 2023-01-25
Refactor driver script [VAN-174]

Technius:
* [x] Drop solang support
* [x] Clean up code structure
* [x] Support multiple input file arguments
* [x] ~~Support multiple detector arguments (this is already supported by the detector program)~~
* [x] Make C++ detector runner handle the "all" argument instead of passing hardcoded list when calling from Python 
* [x] Add `--include-path` option
* [ ] Use `logging` module


Technius:
An output/logging refactor can be left in a separate PR; this one is getting rather large.

vanguard_driver.py
```diff
@@ -192,141 +185,109 @@ def is_cached(out_path: Path, src_path: Path) -> bool:
...
+    if circom2llvm_proc.returncode != 0:
+        print("Failed to run circom2llvm!", file=sys.stdout)
+        print(circom2llvm_proc.stderr, file=sys.stdout)
+        raise RuntimeError("Failed to run circom2llvm")
+    print(' done.', file=sys.stdout)
+    summary_path = args.output_dir / src_path.name.replace(".circom", ".json")
+    return summary_path
+
+
+def run_detector(lang: Platform, detector: str, summary_path: Path, out_dir: Path,
```
kferles, line 281: Why don't we generalize this to multiple detectors? Currently, the python script exposes fewer features than Vanguard's main.

kferles (commented):
Some minor comment but LGTM overall.

vanguard_driver.py
Technius: The detector runner supports multiple detectors with comma separated values. The driver passes the detector argument directly without any changes.

vanguard_driver.py
```diff
@@ -192,141 +185,109 @@ def is_cached(out_path: Path, src_path: Path) -> bool:
...
-            )
-        else:
-            # Directly use output of pyul
-            pyul_cmd = mk_pyul_cmd(args.src_path, args.output_dir, args.base_dir)
-            print("Running pyul...", end="", flush=True, file=sys.stdout)
-            if DEBUG_MODE:
-                print("Pyul cmd: ", *pyul_cmd)
-            pyul_proc = subprocess.run(
-                pyul_cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True
+        if contract_name not in summaries:
```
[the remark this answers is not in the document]
Technius: I think this is a leftover from when the summarizer only accepted one argument rather than the whole solc compiler output. This can probably be removed safely.

vanguard_driver.py
kferles: Oh, ok! All good then!

vanguard_driver.py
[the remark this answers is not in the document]
Technius: done

stephensj2 (approved):

PR #57, opened 2023-01-26
Ben sepanski/yul-llcalls [VAN-163]

benSepanski:
Identify low-level calls from yul2llvm

TODO:
- [x] Figure out regex
<s>- [ ] Test


benSepanski:
@kferles Is there any good way to write tests for Blockchain models? Currently I've just checked it manually on a couple examples in the terminal

benSepanski:
After a conversation with @Technius it seems like tests will be difficult to write at the moment.

Technius:
As discussed in another meeting, one strategy for testing this is to create a "summary debug" detector that prints out what contracts there are, what functions there are in each contract, and what those functions are detected as.

lib/domain/libBlockchain/langs/yul/YulModel.h
```diff
@@ -36,23 +36,21 @@ namespace vanguard {
...
+        bool isLowLevelCall(const CallIns<Domain> &call) const override {
+            // https://github.com/Veridise/yul2llvm/blob/main/lib/libyul2llvm/YulASTVisitor/IntrinsicRewriter.cpp#L71-L126
+            return callNameMatches(call, "pyul_call(_unknown)?");
         }
 
-        bool isLowLevelStaticCall(CallIns<Domain> &call) const override {
-            //throw std::runtime_error("To be implemented");
-            return false;
+        bool isLowLevelStaticCall(const CallIns<Domain> &call) const override {
+            return callNameMatches(call, "delegatecall");
```
stephensj2, line 49: I think Ajinkya has a different designation for the static and delegate calls. We should ask him but I think they are "pyul_delegatecall" and "pyul_staticcall" or something like that

lib/domain/libBlockchain/langs/yul/YulModel.h
```diff
@@ -36,23 +36,21 @@ namespace vanguard {
...
+        bool isAnyLowLevelCall(const CallIns<Domain> &call) const override {
             return isLowLevelCall(call) || isLowLevelStaticCall(call) || isLowLevelDelegateCall(call);
         }
 
-        bool isLowLevelCall(CallIns<Domain> &call) const override {
-            //throw std::runtime_error("To be implemented");
-            return false;
+        bool isLowLevelCall(const CallIns<Domain> &call) const override {
+            // https://github.com/Veridise/yul2llvm/blob/main/lib/libyul2llvm/YulASTVisitor/IntrinsicRewriter.cpp#L71-L126
+            return callNameMatches(call, "pyul_call(_unknown)?");
```
stephensj2, line 45: Based on this definition, there is overlap between isExternalCall and isLowLevelCall. Maybe we should change the api to be "isCall", "isStaticCall" "isDelegateCall". Then change "isExternalCall" to be essentially the current definition of "isAnyLowLevelCall".

lib/domain/libBlockchain/langs/yul/YulModel.h
benSepanski: I'll ask. There's no mention (that I can find) of delegate call explicitly in yul2llvm, so the above regex is based on running pyul on the below example:
```solidity
pragma solidity ^0.8.10;

// NOTE: Deploy this contract first
contract B {
    // NOTE: storage layout must be the same as contract A
    uint public num;
    address public sender;
    uint public value;

    function setVars(uint _num) public payable {
        num = _num;
        sender = msg.sender;
        value = msg.value;
    }
}

contract A {
    uint public num;
    address public sender;
    uint public value;

    function setVars(address _contract, uint _num) public payable {
        // A's storage is set, B is not modified.
        (bool success, bytes memory data) = _contract.delegatecall (
            abi.encodeWithSignature("setVars(uint256)", _num)
        );
    }
}
```
which led to the following LLVM for `A::setVars`:
```
define void @fun_setVars_55(i256 addrspace(1)* %__self, i256 %var__contract_36, i256 %var__num_38) {
entry:
  %alloc_unbounded = alloca i256, align 8
  %alloc_unbounded_casted = ptrtoint i256* %alloc_unbounded to i256
  %0 = add i256 %alloc_unbounded_casted, 32
  %1 = inttoptr i256 %0 to i256*
  store i256 45411953894442575461894080223251671226924700962366181227310544232468229128192, i256* %1, align 4
  %2 = add i256 %0, 4
  %abi_encode_tuple_t_uint256__to_t_uint256__fromStack = call i256 @abi_encode_tuple_t_uint256__to_t_uint256__fromStack(i256 %2, i256 %var__num_38)
  %3 = add i256 %alloc_unbounded_casted, 32
  %4 = sub i256 %abi_encode_tuple_t_uint256__to_t_uint256__fromStack, %3
  %5 = inttoptr i256 %alloc_unbounded_casted to i256*
  store i256 %4, i256* %5, align 4
  %6 = sub i256 %abi_encode_tuple_t_uint256__to_t_uint256__fromStack, %alloc_unbounded_casted
  %finalize_allocation = call i256 @finalize_allocation(i256 %alloc_unbounded_casted, i256 %6)
  %7 = add i256 %alloc_unbounded_casted, 32
  %mload = call i256 @mload(i256 %alloc_unbounded_casted)
  %gas = call i256 @gas()
  %delegatecall = call i256 @delegatecall(i256 %gas, i256 %var__contract_36, i256 %7, i256 %mload, i256 0, i256 0)
  %extract_returndata = call i256 @extract_returndata(i256 addrspace(1)* %__self)
  ret void
}
```

lib/domain/libBlockchain/langs/yul/YulModel.h
benSepanski: Are you referring to the linked definition in yul2llvm?

benSepanski: If so then I think I agree with you. As currently implemented, a low-level call is basically just an external call without a selector.

lib/domain/libBlockchain/langs/yul/YulModel.h
```diff
@@ -36,23 +36,21 @@ namespace vanguard {
...
-        bool isAnyLowLevelCall(CallIns<Domain> &call) const override {
+        bool isAnyLowLevelCall(const CallIns<Domain> &call) const override {
             return isLowLevelCall(call) || isLowLevelStaticCall(call) || isLowLevelDelegateCall(call);
         }
 
-        bool isLowLevelCall(CallIns<Domain> &call) const override {
-            //throw std::runtime_error("To be implemented");
-            return false;
+        bool isLowLevelCall(const CallIns<Domain> &call) const override {
+            // https://github.com/Veridise/yul2llvm/blob/main/lib/libyul2llvm/YulASTVisitor/IntrinsicRewriter.cpp#L71-L126
```
Technius, line 39: This links to `main` branch, so the lines will change; a permalink should be used instead.

lib/domain/libBlockchain/langs/yul/YulModel.h
```diff
@@ -36,23 +36,21 @@ namespace vanguard {
...
-            return false;
+        bool isLowLevelStaticCall(const CallIns<Domain> &call) const override {
+            return callNameMatches(call, "delegatecall");
         }
 
-        bool isLowLevelDelegateCall(CallIns<Domain> &call) const override {
-            //throw std::runtime_error("To be implemented");
-            return false;
+        bool isLowLevelDelegateCall(const CallIns<Domain> &call) const override {
+            return callNameMatches(call, "staticcall");
```
Technius, line 48: This matches against "staticcall" instead of "delegatecall".

I don't think pyul supports delegatecall yet, but it will probably be translated into `pyul_call_delegate` or some similar name.

lib/domain/libBlockchain/langs/yul/YulModel.h
Technius, line 44: This matches against "delegatecall" instead of "staticcall".

Also, note that pyul will translate static calls in a similar manner as normal calls, see https://github.com/Veridise/yul2llvm/pull/79

lib/domain/libBlockchain/langs/yul/YulModel.h
```diff
@@ -189,6 +179,21 @@ namespace vanguard {
             }
         };
 
+        bool callNameMatches(const CallIns<Domain> &call , const std::string &regex) const {
```
Technius, line 176: I think it makes more sense to declare `regex` as `const std::regex&`, since we only use `regex` to construct the `std::regex` instance anyways.
```suggestion
        bool callNameMatches(const CallIns<Domain> &call , const std::regex &reg) const {
```

lib/domain/libBlockchain/langs/yul/YulModel.h
benSepanski: Addressed in https://github.com/Veridise/Vanguard/pull/57/commits/159d11cebffcfb0c31fba33e6253f14c57c39fca

lib/domain/libBlockchain/langs/yul/YulModel.h
benSepanski: Fixed in https://github.com/Veridise/Vanguard/pull/57/commits/ddd3a540c339c36828f0d886ee642f2ff3a63a97

lib/domain/libBlockchain/langs/yul/YulModel.h
benSepanski: Fixed in https://github.com/Veridise/Vanguard/pull/57/commits/de5846dd6fc8fb87d79fecf7021a4cd3bce5c9dd

lib/domain/libBlockchain/langs/yul/YulModel.h
benSepanski: Thanks, for catching that--I've swapped the static vs delegate in https://github.com/Veridise/Vanguard/pull/57/commits/89b60203719412d66e751696468827bd272d19ed

Regarding pyul support for delegate call, I ran the pyul on the example  in https://github.com/Veridise/Vanguard/pull/57#discussion_r1089415022 and it seemed to work (the example is pulled from an online example of invoking a delegate call in solidity). Is that not expected behavior?

benSepanski:
Per several discussions--I've stopped checking for static/delegate calls in https://github.com/Veridise/Vanguard/pull/57/commits/67d99011f3376aaad43b397014cee92da8950f1a and just throw an error instead.

benSepanski:
In light of the above discussions, I've reverted the isStaticCall and isDelegateCalls to just return false in https://github.com/Veridise/Vanguard/pull/57/commits/a49d96287c9c8a0909a0ef28751df8ed2d32981c

lib/domain/libBlockchain/langs/yul/YulModel.h
benSepanski: Now just returns `false` as of https://github.com/Veridise/Vanguard/commit/a49d96287c9c8a0909a0ef28751df8ed2d32981c (see relevant discussion https://github.com/Veridise/Vanguard/pull/57#discussion_r1089433673)

lib/domain/libBlockchain/langs/yul/YulModel.h
benSepanski: I've reverted the `isDelegateCall` to just return false in https://github.com/Veridise/Vanguard/commit/a49d96287c9c8a0909a0ef28751df8ed2d32981c while we wait for pyul support

lib/domain/libBlockchain/langs/yul/YulModel.h
benSepanski: I've reverted the `isStaticCall` to just return false in https://github.com/Veridise/Vanguard/commit/a49d96287c9c8a0909a0ef28751df8ed2d32981c while we wait for pyul support

Technius: For this test case, a future version of pyul will probably return something like `pyul_call_delegate_unknown`.

lib/domain/libBlockchain/langs/yul/YulModel.h
```diff
@@ -34,24 +34,17 @@ namespace vanguard {
...
-        bool isAnyLowLevelCall(CallIns<Domain> &call) const override {
-            return isLowLevelCall(call) || isLowLevelStaticCall(call) || isLowLevelDelegateCall(call);
-        }
 
-        bool isLowLevelCall(CallIns<Domain> &call) const override {
-            //throw std::runtime_error("To be implemented");
-            return false;
+        bool isCall(const CallIns<Domain> &call) const override {
+            // https://github.com/Veridise/yul2llvm/blob/a545bc791de052d0984898fd594846d8e144cdb4/lib/libyul2llvm/YulASTVisitor/IntrinsicRewriter.cpp#L71-L126
+            return callNameMatches(call, std::regex("pyul_call(_unknown|0x[0-9a-fA-F]{8})?"));
```
Technius, line 40: nit: this will compile the regex every time `isCall()` is invoked. We can make it `const static` so that the regex is only compiled once:
```suggestion
            const static std::regex callRegex("pyul_call(_unknown|0x[0-9a-fA-F]{8})?");
            return callNameMatches(call, callRegex);
```

lib/domain/libBlockchain/langs/yul/YulModel.h
```diff
@@ -189,6 +173,20 @@ namespace vanguard {
...
+        bool callNameMatches(const CallIns<Domain> &call , const std::regex &reg) const {
+            auto unwrapped = call.unwrap();
+            if(const llvm::CallInst *unwrappedCall = llvm::dyn_cast<llvm::CallInst>(unwrapped)) {
+                auto calledFn = unwrappedCall->getCalledFunction();
+                if(!calledFn->hasName()) {
+                    return false;
+                }
+
+                std::string name = calledFn->getName().str();
+                return regex_match(calledFn->getName().str(), reg);
```
Technius, line 185: nit: looks like a typo here
```suggestion
                return regex_match(name, reg);
```

Technius (approved):
Code LGTM. Can't test this easily unless we have the "debug summary" detector.

PR #55, opened 2023-01-25
Reentrancy Detector [VAN-118]

benSepanski:
Implement a reentrancy detector

PR #58, opened 2023-01-31
benSepanski/sound-svf-cg  [VAN-168] [VAN-169]

benSepanski:
Creating an intermediate PR to stack the Reentrancy-Detector on (https://github.com/Veridise/Vanguard/pull/55) so that 
* diffs are more manageable
*  I can have people review the API for the SVFCallGraph more easily
* Proper association with Jira tasks


benSepanski:
General design:
* `interproc-analysis/call-graph/CallGraph.h`: interface for interacting with call graphs inside a domain
* `interproc-analysis/call-graph/SVFCallGraph.h`: SVF implementation of `CallGraph`
* `interproc-analysis/call-graph/SVFCallGraphBuilder.h`: Add domain edges to the SVF `SVF::PTACallGraph` so that future analyses using SVF use a sound call graph

lib/analysis/interproc-analysis/SVFUtils.h
```diff
@@ -18,15 +18,29 @@
...
-
         static bool dependsOnTrg(const llvm::Value &src, const llvm::Value &trg);
 
         static std::vector<const llvm::Value*> valueDeps(const llvm::Value *src);
 
+        template<typename Domain>
+        struct CallGraph {
+            static SVF::PTACallGraph *get(const typename Domain::Universe *univ) {
+                initCallGraph();
+                if (!domainEdgesAddedToCallGraph) {
```
kferles, line 35: Can we just do this inside initCallGraph? I see no reason to add a field just for the extra edges.

lib/analysis/interproc-analysis/call-graph/SVFCallGraph.h
```diff
@@ -0,0 +1,138 @@
...
+        }
+
+    private:
+        const Function *src, *dst;
+        const SVF::PTACallGraphEdge * svfEdge;
+
+        std::vector<Instruction*> convertSVFCalls(const SVF::PTACallGraphEdge::CallInstSet &svfCalls) const {
+            std::vector<Instruction*> convertedCalls;
+            for(const SVF::CallICFGNode *callNode:  svfCalls) {
+                const llvm::Instruction *llvmIns = callNode->getCallSite();
```
kferles, line 72: Wouldn't all call nodes of an edge share the same call site? If we want to differentiate between different targets, then we should grab the callee.

lib/analysis/interproc-analysis/call-graph/SVFCallGraphBuilder.h
```diff
@@ -0,0 +1,140 @@
...
+         * Convert domain function to SVF function
+         * @param func the domain function
+         * @return  the corresponding SVF function, creating one if there is none
+         */
+
+        static const SVF::SVFFunction *toSVFFunction(const Function *func) {
+            const llvm::Function *llvmFunction = func->unwrap();
+            auto svfModule = SVF::LLVMModuleSet::getLLVMModuleSet()->getSVFModule();
+            const SVF::SVFFunction *svfFunction = svfModule->getSVFFunction(llvmFunction);
+            if(svfFunction == nullptr) {
```
kferles, line 36: Out of curiosity: what case are you trying to handle here? When does this happen?

lib/analysis/interproc-analysis/call-graph/SVFCallGraphBuilder.h
```diff
@@ -0,0 +1,140 @@
...
+            addMissingEdges(univ, cg, icfg);
+        }
+
+    protected:
+        /**
+         * Get the new edges which should be added to from
+         * @param from the node to start from
+         * @return A list of (callSite, callee) pairs
+         */
+        static std::vector<std::pair<const Function *, const Instruction *>> getNewEdges(Function *from) {
```
kferles, line 94: I don't understand this method. Shouldn't it be using a model to determine additional edges?

kferles (commented):
API looks good to me. I have some minor comments here and there.

lib/analysis/interproc-analysis/call-graph/SVFCallGraph.h
benSepanski: I don't think so. My understanding is that each "call site" is an instruction in the caller which may call the callee, e.g.
```
fun foo():
    x = bar();
    y = bar();

fun bar():
    return 0;
```
there are two different call-sites for the edge `foo -> bar`, namely `x = bar()` and `y = bar()`.

The `ICFGCallNode` class is not a node in the call-graph, but rather a node in the `ICFG` (inter procedural control-flow graph)[ that was used to create the CallGraph](https://github.com/SVF-tools/SVF/blob/e1b9ae9dd6b58a5c9827a8c22f4739ee3c822f40/svf/lib/Util/CallGraphBuilder.cpp#L38-L66).

lib/analysis/interproc-analysis/call-graph/SVFCallGraphBuilder.h
benSepanski: Hmmm... I thought that at one point I was running into this error--but I can no longer reproduce

benSepanski: This used to happen on the `test/reentrancy_test/reentrantx.sol` examples, but no longer does. Maybe a pyul update fixed it? Anyway--I've changed this to a non-nullity assertion in https://github.com/Veridise/Vanguard/pull/58/commits/dde2bc4ad3629d8e118e4a463fc531ad492e1d32

lib/analysis/interproc-analysis/SVFUtils.h
benSepanski: Fixed in https://github.com/Veridise/Vanguard/pull/58/commits/e49faf122f9adaa2aabc86fcd7d9a145c2c891c8

lib/analysis/interproc-analysis/call-graph/SVFCallGraphBuilder.h
kferles: I see. This is very elegant as a solution then.

lib/analysis/interproc-analysis/call-graph/SVFCallGraph.h
kferles: I probably misunderstood the intention here: Does this function simply convert an LLVM instruction to a Domain one? If so, then things look ok.

benSepanski: Yes, that's basically all it's doing

kferles (approved):

PR #59, opened 2023-02-01
Create DebugSummaryDetector

Sapo-Dorado:
Creates a detector to print out useful information about contracts from the summary.


vanguard_driver.py
```diff
@@ -19,7 +19,7 @@
 ############################
 # Detectors
 ############################
-DETECTORS = ["statGen", "fnPrinter", "reentrancy", "irValidator", "uc-outputs", "uc-inputs", "DivideBeforeMultiply", "Flashloan"]
+DETECTORS = ["statGen", "fnPrinter", "reentrancy", "irValidator", "uc-outputs", "uc-inputs", "DivideBeforeMultiply", "Flashloan", "DebugSummary"]
```
Technius, line 22: This doesn't need to be updated anymore.

lib/detectors/DetectorRegistry.h
```diff
@@ -11,6 +11,7 @@
 #include "Reentrancy/ReentrancyDetector.h"
 #include "Statements/DivideBeforeMultiplyDetector.h"
 #include "FlashLoan/FlashloanDetector.h"
+#include "Debug/DebugSummaryDetector.h"
```
Technius, line 14: Is it possible to move this to the cpp file? It's only used there.

Technius (approved):

PR #61, opened 2023-02-02
Create V0 Flashloan Detector

Sapo-Dorado:
Add support for external calls to Flashloan Detector.


test/Flashloan.sol
```diff
@@ -1,13 +1,29 @@
...
-// CHECK: Check the following functions for vulnerabilities:
-// CHECK-DAG: {{vulnerable1}}
-// CHECK-DAG: {{vulnerable2}}
-// CHECK-DAG: {{vulnerable3}}
+// CHECK-DAG: FlashloanVulnerable2: vulnerable
+// CHECK-DAG: vulnerable1
+// CHECK-DAG: vulnerable2
+// CHECK-DAG: vulnerable3
+// CHECK-DAG: FlashloanVulnerable1: vulnerable
+// CHECK-DAG: vulnerable
```
Technius, line 7: This should check that
* FlashloanNotVulnerable is not flagged as vulnerable.
* transfer, balanceOf are not flagged as vulnerable.

lib/domain/libBlockchain/langs/yul/YulModel.h
```diff
@@ -73,9 +73,9 @@ namespace vanguard {
                 }
 
                 std::string name = calledFn->getName().str();
-                std::regex reg("pyul_call_0x[0-9a-fA-F]{8}");
+                std::regex reg("pyul_(static)?call_0x[0-9a-fA-F]{8}");
                 if(regex_match(name, reg)) {
-                    return name.substr(10);
+                    return name.substr(name.size() - 10);
```
Technius, line 78: I think it would be better to use a capture group for this instead of having to update the string every time?

lib/domain/libBlockchain/langs/yul/YulModel.h
```diff
@@ -73,9 +73,9 @@ namespace vanguard {
                 }
 
                 std::string name = calledFn->getName().str();
-                std::regex reg("pyul_call_0x[0-9a-fA-F]{8}");
+                std::regex reg("pyul_(static)?call_0x[0-9a-fA-F]{8}");
```
Technius, line 76: This is the same regex as the one above. I think it would make sense to refactor it into a constant at the top of the file.

lib/domain/libBlockchain/langs/yul/YulModel.h
```diff
@@ -37,7 +37,7 @@ namespace vanguard {
 
         bool isCall(const CallIns<Domain> &call) const override {
             // https://github.com/Veridise/yul2llvm/blob/a545bc791de052d0984898fd594846d8e144cdb4/lib/libyul2llvm/YulASTVisitor/IntrinsicRewriter.cpp#L71-L126
-            const static std::regex callRegex("pyul_call(_unknown|_0x[0-9a-fA-F]{8})?");
+            const static std::regex callRegex("pyul_(static)?call(_unknown|_0x[0-9a-fA-F]{8})?");
```
stephensj2, line 40: I thought this was included in Ben's refactoring of the external call stuff but it appears not to have been. You've captured the static call case in with the low-level call case here. We want to keep these two things separate rather than catching both with this API (As static calls revert on a variable modification). To do so, keep the original regex then add `callRegex("pyul_(static)?call(_unknown|_0x[0-9a-fA-F]{8})?");` as the static call Regex. You can keep the one to resolve selectors the same.

Sapo-Dorado: Updated this

stephensj2 (approved):

PR #67, opened 2023-02-07
Refactor driver to use logging

Technius:
This refactors the driver to use the Python logging module to handle log output.


kferles (approved):
LGTM

vanguard_driver.py
```diff
@@ -113,6 +120,9 @@ class VanguardDriverOpts:
     # Output format
     sarif: Optional[Path]
 
+    # Logger
+    logger: logging.Logger
```
0xddom, line 124: Why not use a global object and configure it in the main method?

vanguard_driver.py
```diff
@@ -290,24 +299,25 @@ def run_detector(lang: Platform, detector: str, summary_path: Path, out_dir: Pat
...
-    if verbose:
-        print(proc.stderr, file=sys.stderr)
+    logger.info("Completed Vanguard.")
+    for line in proc.stderr.splitlines():
+        logger.debug(proc.stderr)
 
     if extra_sarif_path:
-        print(f"SARIF results are available at: {extra_sarif_path}", file=sys.stderr)
+        logger.info(f"SARIF results are available at: {extra_sarif_path}")
         shutil.copy(sarif_artifact, extra_sarif_path)
```
0xddom, line 321: Why not make Vanguard write directly to `extra_sarif_path`?

vanguard_driver.py
```diff
@@ -378,12 +389,29 @@ def run(args: VanguardDriverOpts, lang: Platform):
...
+            args.logger.info(line)
+
+
+def _configure_logging(logger: logging.Logger, log_level: int):
+    logger.setLevel(log_level)
+    # WARNING severity and above is logged to stderr.
+    err_handler = logging.StreamHandler(stream=sys.stderr)
+    err_handler.setLevel(logging.WARNING)
+    err_handler.setFormatter(
+        logging.Formatter(fmt="{levelname}: {message}", style="{"))
```
0xddom, line 403:
```suggestion
        logging.Formatter(fmt="{asctime} [{levelname}]: {message}", style="{"))
```

IMO always good to have the date in a log

0xddom:
wrt circom2llvm lgtm

vanguard_driver.py
Technius: Makes it harder to refactor if we want to change the logger hierarchy later down the line.

vanguard_driver.py
Technius: That's a good idea, but it's out-of-scope for this PR.

vanguard_driver.py
Technius: I generally agree with you, but the current mode is meant for interactive command line use; timestamps will clutter up the log. There's nothing preventing the user from attaching timestamps by [piping to external tools](https://unix.stackexchange.com/questions/26728/prepending-a-timestamp-to-each-line-of-output-from-a-command), so this is a matter of whether we want to show the timestamp by default.

I'll soon be submitting a follow up PR that has the json logging mode. As the json mode is not meant for human consumption, it includes more details like timestamps and source location by default.

0xddom: I see, I was not aware of that. In that case makes sense to do it that way

PR #68, opened 2023-02-08
Create V1 Flashloan Detector 

Sapo-Dorado:
Update V0 Flashloan Detector to track tainting contract variables.


test/Flashloan.sol
```diff
@@ -1,129 +1,98 @@
...
-// CHECK-DAG: FlashloanVulnerable2: vulnerable
-// CHECK-DAG: vulnerable1
-// CHECK-DAG: vulnerable2
-// CHECK-DAG: vulnerable3
-// CHECK-DAG: FlashloanVulnerable1: vulnerable
+// RUN: vanguard_driver --detector flashloan -D %S --src_path %s | FileCheck %s \
+// RUN: --implicit-check-not {{[nN]otVulnerable}} --implicit-check-not transfer --implicit-check-not balanceOf \
+// RUN: --implicit-check-not taintVariable --implicit-check-not deposit
+// CHECK-DAG: SimulatedTokenBalanceFlashloan: vulnerable
+// CHECK-DAG: steal
```
Technius, line 6: Can we improve the report here? This is not very user-friendly. I would suggest the following as an initial version of the report:
```
Potential flashloan issue detected in MyContract.myMethod:
  - ExampleToken.balanceOf() is used to calculate amount transferred in call to ExampleToken.transferFrom()
```

Later, we can further augment a report like this with line numbers.

Of course, it would be great if you have more ideas for how we can improve the report.

PR #70, opened 2023-02-09
Add a --list-detectors option to list detectors [VAN-197]

Technius:
* Add a `--vanguard-list-detectors` flag to the detector runner to list all registered detectors.

* Add a `--list-detectors` flag in the driver that calls detector runner with above.


Technius:
Build failure is a cachix problem, this should be fine to merge.

PR #71, opened 2023-02-09
SVF Patch to get rid of compiler warning

benSepanski:
SVF's SparseBitVector.h uses an `assert(False,...)` to guard a control flow path that should never be taken. This causes a compiler warning that not all control-flow paths have a return statement whenever the file is included. 

This PR adds a patch to the SVF nix build to change the assert to a throw so that this warning goes away.


Technius (approved):

Technius:
The build log shows a similar warning here:
```
svf> /build/source/include/Util/SparseBitVector.h: In member function 'int SVF::SparseBitVectorElement<ElementSize>::find_last() const [with unsigned int ElementSize = 128]':
```
We don't seem to use this, so maybe it's fine to leave the other one as-is for now.

benSepanski:
Ah sorry--I merged right before I saw the comment. I guess we'll just leave it as-is for now, and add it to the patch if it comes up later

PR #72, opened 2023-02-10
Add support for JSON logging

Technius:
See the commit descriptions for details.

NOTE: this may break some development workflows, as one more step is needed: installing the Python dependencies.


kferles (approved):
LGTM. Although I am not familiar with the library.

Technius:
I added tests that lightly check the json output stuff. The `vanguard_driver` command can also be directly invoked in the test suite now, and the tests _will_ break if your environment is stale.

PR #76, opened 2023-02-15
Adding a fix for calculating R/W sets with multiple source files

stephensj2:


benSepanski:
Any chance we could also add a non-empty assertion inside the `writesStorage(ins)` case of `YulModel::getStorageWrites(...)`?

PR #77, opened 2023-02-16
Ben sepanski/flash minter [VAN-194]

benSepanski:
Incorporates flash minter into the reentrancy tests, fixes bug so correct reentrancies are reported, and applies fixes to improve presentation

PR #78, opened 2023-02-17
Ben sepanski/include interface summaries [VAN-195]

benSepanski:
Stacked onto https://github.com/Veridise/Vanguard/pull/77 to keep PRs small.

Includes interface summaries during the driver script, and makes appropriate fixes for the Reentrancy Detector.


vanguard_driver.py
```diff
@@ -240,6 +240,13 @@ def generate_solidity_llvm(args: VanguardDriverOpts) -> Path:
         cdata["name"]: cdata for cdata in summary_dict["contracts"]
     }
 
+    # Prepare summary for interfaces
+    def is_interface(summary: dict) -> bool:
+        return summary.get("kind") == "interface"
+
+    interface_summaries: List[dict] = [summary for summary in summaries.values() if is_interface(summary)]
+    interfaces_included = len(interface_summaries) == 0
```
Technius, line 248: Why is this needed? Why not always include the interfaces?

vanguard_driver.py
```diff
@@ -255,15 +262,25 @@ def generate_solidity_llvm(args: VanguardDriverOpts) -> Path:
         # Copy .ll file to make it easier to work with
         shutil.copy(ll_file, ll_in_path, follow_symlinks=True)
 
+        # pretty-print if debugging
+        indent = None
+        if logging.getLogger().isEnabledFor(logging.DEBUG):
```
Technius, line 267: Might be better to check against `args.log_level` instead of inspecting the global root logger.

vanguard_driver.py
```diff
@@ -255,15 +262,25 @@ def generate_solidity_llvm(args: VanguardDriverOpts) -> Path:
...
+            indent = "  "
+
         # Copy summary data from summary json
+        contracts = [summaries[contract_name]]
+        if not interfaces_included:
+            contracts.extend(interface_summaries)
```
Technius, line 273: This only appears to affect the first contract that is found.

vanguard_driver.py
benSepanski: It's not necessary, it just sets `interfaces_included` to true when there are no interface summaries to include. I can change it to `False` for code clarity

vanguard_driver.py
benSepanski: Yes, Vanguard has no checks for duplicate contract IDs. If I include the interface into multiple contracts then duplicate Contract objects are created in the universe

vanguard_driver.py
benSepanski: Addressed in https://github.com/Veridise/Vanguard/pull/78/commits/088b4aba96c31da663402f1d97c631f863d16dda (used `args.logger` here since the referenced `args` is a `VanguardDriverOpts`)

vanguard_driver.py
Technius: Seems like we should fix this issue in the detector program and not the driver, then. Let's fix it in a follow-up PR so we don't block this PR.

benSepanski: SGTM

lib/detectors/CallGraphReentrancy/ReentrancyReportFormatter.h
```diff
@@ -153,8 +153,11 @@ namespace vanguard {
 
         ReentrancyReportFormatter& function(const typename Domain::Function *func) {
             assert(func != nullptr);
-            this->recordLocation(func->name());
-            return *this << func->name();
+            std::string name = func->name();
+            if(func->contract() != nullptr) {
+                name = func->contract()->name() + "." + name;
+            }
+            this->recordLocation(name); return *this << name;
```
Technius, line 160: nit: need a newline here

Technius (approved):
LGTM.

lib/detectors/CallGraphReentrancy/ReentrancyReportFormatter.h
benSepanski: Fixed in https://github.com/Veridise/Vanguard/pull/78/commits/80b942229df799ae12de524841e3b78fd97e9abb

lib/analysis/interproc-analysis/call-graph/SVFCallGraphBuilder.h
```diff
@@ -104,14 +121,22 @@ namespace vanguard {
             }
         };
 
+        static std::shared_ptr<SVF::PTACallGraphNode> externalNode;
```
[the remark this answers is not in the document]
benSepanski: Currently, all methods of SVFCallGraphBuilder are static, so the class is never actually instantiated.

PR #80, opened 2023-02-20
Taint Analysis ❤️ SVFUtils

0xddom:


lib/analysis/interproc-analysis/SVFUtils.cpp
```diff
@@ -75,7 +76,11 @@ namespace vanguard {
         if (svfg == nullptr) {
             // Ensure pa is initialized.
             getPointerAnalysis();
+#if 0
```
kferles, line 79: Oh, these seem to be two different SVF-internal builders. Can we just create a `getTaintSVFG` which initializes and returns a different SVFG?

kferles (changes requested):

lib/analysis/interproc-analysis/SVFUtils.cpp
0xddom: No, `TaintSVFGBuilder` is in SVF's namespace because it needs to be a friend of SVFG, but it is my custom class.

kferles: I see. Ok, I think we can still do what I suggested above.

0xddom: But wasnt the point of moving the taint analysis svfg to utils for sharing? If we maintain two pointers changes in one SVFG will not replicate to the other

0xddom:
Proposal: Refactor the TaintAnalysis module to use dependency injection, leaving to the user the decision of what SVFG to use. Also add a decorator that adds the TaintAnalysis specific modifications (globals and stuff) in case the user of the TaintAnalysis class needs them. If we run multiple detectors we may contaminate the global SVFG unless the decorator uses some kind of memento to rollback changes

lib/analysis/interproc-analysis/SVFUtils.cpp
kferles: Ok, let's merge it as is for now and we can decide late how to resolve this issue. I am not sure I understand the actual issue.

kferles (approved):

PR #81, opened 2023-02-20
Clean up driver script

Technius:


Technius:
@benSepanski I think this will introduce some small conflicts with #77. Can you see if anything will break?

test/Inputs/PuppetPool.sol
```diff
@@ -1,4 +1,4 @@
-// RUN: python3 vanguard_driver.py --src_path %s --detector statGen
+// RUN: python3 vanguard_driver.py --detector statGen
```
benSepanski, line 1: Should this use the `vanguard_driver` command?
```suggestion
// RUN: vanguard_driver --detector statGen
```

benSepanski (approved):
Looks good to me. I don't think the changes will be break https://github.com/Veridise/Vanguard/pull/77.

Do you want me to wait to merge https://github.com/Veridise/Vanguard/pull/77 and I can resolve the conflicts? If not I'll go ahead and merge it now

test/Inputs/PuppetPool.sol
Technius: Yes, you're right. However, the PuppetPool test case actually is run through a `.txt` file so this line should actually be removed.

Technius:
if #77 is ready then it would be better to merge #77 first and then rebase this one on top.

PR #83, opened 2023-02-20
Distinguish application from library code [VAN-198]

benSepanski:
Application directories are those supplied to the vanguard driver as "source paths" (or the enclosing directory of files passed as source paths.

A module is an application module if both of the following hold:
* its source file exists (this is necessary because the source paths stored in the modules are relative)
* its source file is located inside of an application directory


lib/Vanguard.cpp
```diff
@@ -221,6 +227,20 @@ static void VanguardMain(int argc, char **argv, std::vector<std::string> &errors
...
 
     initializeLLVM(argc, argv);
+    // Convert application directories to paths and fail now if they don't exist
+    auto absAppDirs = std::make_shared<std::vector<std::string>>();
+    auto toAbs = [](const std::string &appDir) -> std::string {
+        llvm::SmallString<128> absAppDir = llvm::StringRef(appDir);
+        llvm::sys::fs::make_absolute(absAppDir);
+        if(!llvm::sys::fs::is_directory(absAppDir)) {
+            llvm::WithColor::error() << "application directory " << absAppDir << " does not exist!\n";
+            throw std::runtime_error("Application directory does not exist");
```
Technius, line 237:
```suggestion
            llvm::WithColor::error() << "application directory " << absAppDir << " is not a directory!\n";
            throw std::runtime_error("Application directory is not a directory");
```

lib/Vanguard.cpp
```diff
@@ -221,6 +227,20 @@ static void VanguardMain(int argc, char **argv, std::vector<std::string> &errors
...
+        }
+        llvm::outs() << static_cast<std::string>(absAppDir) << "\n";
```
Technius, line 239: This looks like a stray debug line

vanguard_driver.py
```diff
@@ -377,21 +382,27 @@ def run(args: VanguardDriverOpts, lang: Platform):
 
     args.logger.info(f"----Running Vanguard with {args.detector} detector----")
 
+    to_dir: Callable[[Path], Path] = lambda path: path if path.is_dir() else path.parent
+    get_app_dir: Callable[[Path], str] = lambda path: str(to_dir(path).absolute())
+    application_dirs: List[str] = list(map(get_app_dir, args.src_paths))
```
Technius, line 387: nit: `get_app_dir` seems unnecessarily verbose to me, I would simplify this to
```suggestion
    to_dir: Callable[[Path], Path] = lambda path: path if path.is_dir() else path.parent
    application_dirs: List[str] = [str(to_dir(path).absolute()) for path in args.src_paths]
```
Also, since the `application_dirs` are only used to build the Vanguard command, it's fine to leave them as `Path`s.

Technius (approved):

lib/Vanguard.cpp
benSepanski: Ack--thanks for catching! Fixed in https://github.com/Veridise/Vanguard/pull/83/commits/427eadff93d313465f0ffaae46e708b4fb6bdaf3

vanguard_driver.py
benSepanski: Simplified in https://github.com/Veridise/Vanguard/pull/83/commits/ee87a6f9fc2bb646ba0ec6318f5bf5b37b277e48

Technius:
This needs to be rebased.

benSepanski:
> This needs to be rebased.

Done. Currently this is failing b/c of flash minter. I am creating a benSepanski/ethdenver branch in yul2llvm which should be sufficient for the demo. With that done, I'll put that in the flake.lock for Vanguard

benSepanski:
@Technius marking this as ready for review again now that the yul2llvm issue has been resolved. Only changes have been [these last two commits](https://github.com/Veridise/Vanguard/pull/83/files/c76324510a860a744ac5386ac67b68e3a0a19117..190eb2a86d01c55fad7b07504a422a63537bce83) and your [suggested fixes](https://github.com/Veridise/Vanguard/pull/83/files/a3697821ad4026e7c681981a451630f54bbf8d61..c76324510a860a744ac5386ac67b68e3a0a19117)

PR #85, opened 2023-02-22
Handle case where only base dir passed

benSepanski:
This fixes the previously unaccounted-for case in which `base-dir` is passed, but no sources are passed


test/debug-summary-tests/ApplicationDependency.sol
```diff
@@ -1,10 +1,14 @@
...
-// RUN: FileCheck %s --check-prefixes APP --input-file %t
-// RUN: FileCheck %s --check-prefixes DEP --input-file %t
+
+// RUN: for file_args in "-D %S %S/application-code" "-D %S/application-code --include-dir %S -L solidity" ; do \
+// RUN:      echo ${file_args} ; \
+// RUN:      vanguard_driver --detector DebugSummary ${file_args} > %t ; \
+// RUN:      FileCheck %s --check-prefixes CHECK --input-file %t ; \
+// RUN:      FileCheck %s --check-prefixes APP --input-file %t ; \
+// RUN:      FileCheck %s --check-prefixes DEP --input-file %t ; \
+// RUN: done
```
benSepanski, line 11: This would be prettier using the DEFINE/REDEFINE directives described here: https://reviews.llvm.org/D132513, but a search on search.nixos.org indicates the only available version of lit is 14.0. I think we would need 15.0.7 (maybe 15.0.*, I'm not exactly sure)

benSepanski: @Technius any chance we could pip install lit (post-eth Denver) rather than grabbing the [Nixpkgs version](https://search.nixos.org/packages?channel=22.11&show=lit&from=0&size=50&sort=relevance&type=packages&query=lit)?

kferles (approved):
LGTM

PR #86, opened 2023-02-23
update DivideBeforeMultiplyDetector

Sapo-Dorado:
Fixes a bug in the DivideBeforeMultiplyDetector that resulted in false positive outputs for not vulnerable functions that contained a division operation in vulnerable contracts.

Updates the name of the detector to `divide-before-multiply` to adhere to naming conventions

Improves the output to be more user friendly by explaining the issue


lib/detectors/Statements/DivideBeforeMultiplyDetector.h
```diff
@@ -21,41 +21,62 @@ namespace vanguard {
...
 
-        struct CallTargetResolver : public InstructionClassVisitor<Domain> {
+        struct MulDivResolver : public InstructionClassVisitor<Domain> {
             void visit(const CallIns<Domain> &v) override{
-                tgts = v.targets();
+                for(auto tgt : v.targets()) {
+                    if(tgt->isHelper()){
+                        for (auto blk_inner: tgt->blocks()) {
+                            for(auto ins_inner: blk_inner->instructions()) {
+                                ins_inner->accept(*this);
```
Technius, line 30: Will this recurse infinitely if there are mutually recursive functions? Or do helper functions not recurse in practice?

Technius (approved):
LGTM. A few minor comments.

Technius:
The analysis seems somewhat coarse-grained but I think this is fine as an initial version. We can improve the false positive rate later. I think getting precise information about where the issue happens would be more useful at the moment...

lib/detectors/Statements/DivideBeforeMultiplyDetector.h
Sapo-Dorado: I'm not sure, that might be an issue. I only need to go one level deep in examining the targets to evaluate the multiply and divide operations so I will change this

lib/detectors/Statements/DivideBeforeMultiplyDetector.h
```diff
@@ -64,31 +85,25 @@ namespace vanguard {
...
-            std::unordered_set<std::string> potentiallyVulnerable;
-            std::unordered_set<std::string> contractFns;
+            std::unordered_set<std::string> &vulnerableFns;
+            std::unordered_set<Function *> &functions;
         };
 
 
 
-        void process(Function *fn, MulDivResolver *mulDivResolver) {
+        void process(Function *fn) {
```
Technius, line 96: This will segfault on functions with no body/implementation (e.g., if interfaces exist). This issue also affects DebugSummary, and I'm not sure how to solve it--the underlying LLVM function will be null in this case. @benSepanski do you know how to handle interface functions?

Technius (changes requested):

lib/detectors/Statements/DivideBeforeMultiplyDetector.h
Technius: nvm, I think we can check `fn->unwrap()` and do nothing if it is null.

benSepanski: There are basically two options to check if the function is an interface function.
1. What I did in [SVFCallGraphBuilder](https://github.com/Veridise/Vanguard/blob/02915c7f13a629eebf9dcad4b45c8543625ba7c3/lib/analysis/interproc-analysis/call-graph/SVFCallGraphBuilder.h#L37-L42) is to just check if `function->unwrap() == nullptr`. I didn't realize it at the time, but this is functionally checking whether or not `function` is an interface function.
2. Probably a better way to handle this is to check `fn->contract()->kind() == ContractKind::INTERFACE`

Technius: Got it. I think we could clean up the API here by introducing a `function->hasDefinition()` (so we don't have to worry about interfaces or abstract functions -- and this should also work across multiple languages). It might also be useful to change `body()` or `instructions()` to return an empty vector if there is no definition, but this has a small tradeoff (no warning/error/failure when unexpectedly operating over undefined functions).

Technius: The API changes I mentioned above is out of scope of the PR, but we should do it in the near future. For this PR, one of the checks that @benSepanski mentioned above would be appropriate.

benSepanski: There is something like that, [`function->hasBody()`](https://github.com/Veridise/Vanguard/blob/02915c7f13a629eebf9dcad4b45c8543625ba7c3/lib/program/Function.h#L33-L35). I forgot about that function while writing the earlier response

benSepanski: See usage in e.g. [CrossContractCallClassifier](https://github.com/Veridise/Vanguard/blob/02915c7f13a629eebf9dcad4b45c8543625ba7c3/lib/detectors/CallGraphReentrancy/CrossContractCallClassifier.h#L83)

Technius: [function->hasBody()](https://github.com/Veridise/Vanguard/blob/02915c7f13a629eebf9dcad4b45c8543625ba7c3/lib/program/Function.h#L33-L35) unfortunately calls a method on the underlying LLVM function, which won't work if the underlying function is null...

benSepanski: Ack--god point

PR #89, opened 2023-03-02
Improve handling of interface functions in SVF call graph wrapper

benSepanski:
@Sapo-Dorado Here's a fix for the issue. Passes all the lit tests, lmk if it has any issues

Basically I had been handling interface functions by just making a dummy node for them. This whole process can be avoided by just instead not calling the SVF Call Graph on interface contracts. Since this seems to be what people are doing in practice, I just decided to assert that they don't call SVF call graph functions on interfaces


benSepanski:
Okay should be ready now

PR #91, opened 2023-03-11
Add some extra debug info

Technius:
This should be helpful in debugging some unexplainable segfaults we're getting when running in Docker.


lib/detectors/Debug/DebugSummaryDetector.h
```diff
@@ -101,6 +131,9 @@ namespace vanguard {
                 info.contractNameToIsApplicationCode[currContract] = isApp;
                 for(auto *fn: contract->functions()) {
                     info.numFunctions++;
+                    // TODO: need a saner way to loop over functions that won't crash
+                    if (!fn->unwrap())
+                        continue;
```
Sapo-Dorado, line 136: Should we add a check like this to other detectors like flashloan and divide before multiply because they suffer from the same issue?

benSepanski:
This looks good to me overall. Seems like the [ApplicationDependency](https://github.com/Veridise/Vanguard/blob/a6e080c0676c44a061c934256372aff9d3cf23bd/test/debug-summary-tests/ApplicationDependency.sol) test needs to be adjusted for the new output from DebugSummary

lib/detectors/Debug/DebugSummaryDetector.h
benSepanski: We probably should eventually. IMO it might be worth changing some of the Function APIs to return std::optional<> (e.g. `unwrap() -> optional<llvm::Function*>`, `uncheckedUnwrap() -> llvm::Function*`) so that these checks are forced upon the detector writers

benSepanski (commented):
This looks good to me overall. Seems like the [ApplicationDependency](https://github.com/Veridise/Vanguard/blob/a6e080c0676c44a061c934256372aff9d3cf23bd/test/debug-summary-tests/ApplicationDependency.sol) test needs to be adjusted for the new output from DebugSummary

lib/detectors/Debug/DebugSummaryDetector.h
Technius: Ideally, we also shouldn't have to unwrap the LLVM function. The generic function API should allow us to query common information without crashing.

PR #92, opened 2023-04-07
[WIP] Share a unified summary for all modules

Technius:
This refactors the blockchain SummaryReader so that all of the summary information can be used for each LLVM module, without resulting in weird duplication or other problems.


lib/domain/libBlockchain/Factory.h
```diff
@@ -19,13 +19,13 @@ namespace vanguard {
     template<typename Base, typename Domain>
     class Blockchain<Base, Domain>::Factory : public Base::Factory {
     public:
-        typename Domain::Function *createBlkFn(const llvm::Module &module, BlockchainModel<Domain> &model, std::string contractName,
+        typename Domain::Function *createBlkFn(const llvm::Module *module, BlockchainModel<Domain> &model, std::string contractName,
```
benSepanski, line 22: Any chance we can add documentation about how the `module==nullptr` case is handled?

lib/Vanguard.cpp
```diff
@@ -339,10 +338,14 @@ static void VanguardMain(int argc, char **argv, std::vector<std::string> &errors
...
         for(auto &mod : modules) {
-            auto sumReader = loadSummary<vanguard::SummaryReader>(mod, filenames[mod]);
-            units.push_back(factory.createBlkUnit(mod, sumReader->contracts()));
-
+            sumReader->linkModule(mod);
+        }
+        // TODO: pass a single summary, and omit the initial module needed.
+        loadSummary(*sumReader, filenames[modules.front()]);
+        for (const auto &[mod, contract] : sumReader->contracts()) {
+            units.push_back(factory.createBlkUnit(mod, {contract}));
```
benSepanski, line 348: Why is there a 1-1 module-contract relationship? I thought each unit could contain multiple contracts?

lib/domain/libBlockchain/SummaryReader.h
```diff
@@ -25,25 +25,33 @@ namespace vanguard {
...
     public:
 
-        SummaryReader(const llvm::Module &mod, const std::string& projectFile);
+        SummaryReader(llvm::LLVMContext *context);
+
+        /// Load the summary file.
+        void load(const std::string &summaryFile);
+
+        /// Provide the given LLVM module for linkage at summary loading time.
+        /// \param mod The LLVM module to link with a contract in the summary.
```
benSepanski, line 37: This is picky--but is our doxygen set up to generate documentation for triple-backshlash function comments?

lib/domain/libBlockchain/SummaryReader.h
```diff
@@ -52,40 +60,40 @@ namespace vanguard {
...
             }
         }
 
-        ContractKind readContractKind(const std::string& kindString);
+        static ContractKind readContractKind(const std::string& kindString);
 
         void readSummary(rapidjson::Value &val);
 
-        BlockchainDomain::Contract *readContract(rapidjson::Value &val);
+        std::pair<BlockchainDomain::Contract*, llvm::Module*> readContract(rapidjson::Value &val);
```
benSepanski, line 77: Should we add `#include <utility>` to help IDEs find `std::pair`?

lib/domain/libBlockchain/langs/LanguageFactory.h
```diff
@@ -25,12 +25,16 @@ namespace vanguard {
 
         // TODO: Reduce number of arguments in the below methods.
 
-        virtual BlockchainDomain::Function *createBlkFn(rapidjson::Value &val, const llvm::Module &module, BlockchainModel<BlockchainDomain> &model, std::string contract,
+        virtual BlockchainDomain::Function *createBlkFn(rapidjson::Value &val, const llvm::Module *module, BlockchainModel<BlockchainDomain> &model, std::string contract,
```
benSepanski, line 28: Not related to this PR, but is this class missing a virtual destructor?

benSepanski (commented):
Overall this looks good to me. I added a couple minor comments

Do we have any explicit guidelines on when a module will be null? Or is that a solidity front-end question

lib/domain/libBlockchain/Factory.h
```diff
@@ -19,13 +19,13 @@ namespace vanguard {
...
                                                ContractKind kind, std::string name, std::string selector, bool isCnstr, Visibility vis, Mutability mut,
                                                std::vector<vanguard::Variable<Domain> *> params, std::vector<vanguard::Variable<Domain> *> rets, std::vector<std::string> mods) {
             auto fn = new typename Domain::Function(name, selector, isCnstr, vis, mut, std::move(params), std::move(rets), std::move(mods), nullptr);
 
-            if(kind != INTERFACE) {
-                for (auto &llvmFn: module) {
+            if (module) {
```
kferles, line 27: Does module != null imply that kind != INTERFACE?

kferles (approved):
I agree with BenS comments. Other than a minor comment LGTM.

lib/domain/libBlockchain/Factory.h
Technius: Yeah, no code will be generated for an interface, and hence there is no module available for an interface.

lib/Vanguard.cpp
Technius: The Yul frontend will generate a separate LLVM module for each contract. I'll add a comment noting this.

lib/domain/libBlockchain/langs/LanguageFactory.h
Technius: Indeed, will fix this too.

lib/domain/libBlockchain/SummaryReader.h
Technius: Yes, this is enabled by the `JAVADOC_AUTOBRIEF` that is set in `CMakeLists.txt`.

lib/domain/libBlockchain/SummaryReader.h
Technius: It works in my LSP setup. Does it show as a missing definition on your computer?

lib/domain/libBlockchain/Factory.h
Technius: Done

lib/Vanguard.cpp
Technius: Done

lib/domain/libBlockchain/langs/LanguageFactory.h
Technius: Done

lib/Vanguard.cpp
benSepanski: I guess the `<utility>` include is unnecessary. Marking as resolved :)

Technius:
I squashed the changes so this should be ready to merge.

There's one last bit that's bothering me: do we want to assume that every interface and contract has an `llvm::Module`, or should we allow them to be null? Currently, the code will generate an empty module for contracts that do not have concrete implementations.

lib/Vanguard.cpp
Technius: One more note: for other languages (e.g., NEAR), there may be different assumptions about what how modules relate to contracts. The summary reader will need some slight, but not major, modifications to get it to work with non-Solidity languages. E.g., if a module may contain multiple contracts, the psuedocode will look more like "for every module M, create a compilation unit where M is mapped to all contracts implemented in M."

PR #93, opened 2023-04-11
Create Unchecked Return Detector

Sapo-Dorado:
Detects return values from internal, external, or low level calls that aren't used. Will also support unused return values of library calls once Vanguard properly handles library calls.


test/Inputs/SafeMath.sol
```diff
@@ -0,0 +1,215 @@
+// SPDX-License-Identifier: MIT
```
Technius, line 1: Is this a copy of OZ's SafeMath? The test environment should have a copy of OZ that can be passed as an include path to solc.

lib/detectors/Operations/UncheckedReturnDetector.h
```diff
@@ -0,0 +1,104 @@
...
+                        }
+                    }
+                }
+            }
+        }
+
+        void detect(typename Domain::Universe &universe) override {
+            auto contracts = universe.contracts();
+            for(auto contract : contracts) {
+                if(contract->kind() == CONTRACT) {
```
Technius, line 69: Would it make sense to also check libraries?

lib/detectors/Operations/UncheckedReturnDetector.h
```diff
@@ -0,0 +1,104 @@
...
+                    if(ins.operandAt(i) == value) {
+                        return true;
+                    }
+                }
+                return false;
+            }
+            Value* value;
+        };
+
+        void process(Function *fn) {
```
Technius, line 48: Need to skip the function if `fn->unwrap()` is null, or else we'll segfault when calling `fn->blocks()`.

We should definitely fix the program module so that it doesn't break randomly like this.

lib/detectors/Operations/UncheckedReturnDetector.h
```diff
@@ -0,0 +1,104 @@
+//
+// Created by Nicholas Brown on 1/9/23.
+//
+
+#ifndef VANGUARD_UNCHECKEDRETURNDETECTOR_H
+#define VANGUARD_UNCHECKEDRETURNDETECTOR_H
+
+#include<unordered_map>
```
Technius, line 8: A few includes are missing here, which will break IDE autocompletion. This can be debugged by temporarily adding a cpp file that includes this header.

Technius (changes requested):

lib/detectors/Operations/UncheckedReturnDetector.h
Sapo-Dorado: fixed in `detect()`

lib/detectors/Operations/UncheckedReturnDetector.h
```diff
@@ -0,0 +1,111 @@
...
+            }
+        }
+
+        void detect(typename Domain::Universe &universe) override {
+            auto contracts = universe.contracts();
+            for(auto contract : contracts) {
+                if(contract->kind() != INTERFACE) {
+                    for (auto fn: contract->functions()) {
+                        if(fn->unwrap()) {
+                            process(fn);
```
benSepanski, line 77: Any chance this can be implemented as a `FunctionDetector`?

lib/detectors/Operations/UncheckedReturnDetector.h
```diff
@@ -0,0 +1,111 @@
...
+        using Value = typename Domain::Value;
+
+        struct CallInsResolver : public InstructionClassVisitor<Domain> {
+             void visit(const CallIns<Domain> &v) override {
+                 isCall = true;
+                 result = v.result();
+                 if(v.targets().size() == 1) {
+                     target = v.targets().front()->name();
+                 } else {
+                     target = "an unspecified location";
```
benSepanski, line 31: If there are more than 1 targets, why do we just say "unspecified" rather than listing the targets?

benSepanski (commented):
Looks great. I had two small questions, but otherwise lgtm

lib/detectors/Operations/UncheckedReturnDetector.h
benSepanski: TBH, I think our current handling of modules/functions as possibly null ought to be rethought.

@Technius I'd prefer that whenever we have a pointer in Vanguard, it is expected to be non-null. Maybe we can find a time to discuss re-designing the code base to rely on `optional`s for interface/abstract class handling?

lib/detectors/Operations/UncheckedReturnDetector.h
Sapo-Dorado: I wrote it this way because calls like the following test case make a call to an arbitrary address, so the targets are all of the available functions, which would spam the reporting and not reflect the fact that the target can be outside of the project.
```
function unchecked_call(address payable dst) public payable{
       dst.call{value:msg.value}("");
}
```
I did not consider the case where a call has multiple targets that are all relevant, so if these are important enough we can list all the targets.

lib/detectors/Operations/UncheckedReturnDetector.h
Technius: Yes, let's discuss this synchronously.

Technius (approved):

lib/detectors/Operations/UncheckedReturnDetector.h
benSepanski: Seems like a reasonable decision to me, we can change later if necessary

benSepanski (approved):

PR #94, opened 2023-04-11
Some small misc utilities

Technius:
* Add assertion macros
* Change crash dump message


lib/VanguardSupport.h
```diff
@@ -0,0 +1,31 @@
...
+        std::cerr << msg << " at " << filename << ":" << lineno << "\n";
+        std::abort();
+    }
+}
+
+[[noreturn]] inline void __vanguard_todo_impl(bool expr, const char *msg,
+                                              const char *filename,
+                                              const int lineno) {
+    __vanguard_assert_impl(false, msg, filename, lineno);
+    // silence "function declared noreturn should not return" error
```
benSepanski, line 26: Is this a leftover TODO from before `[[noreturn]]` was added, or is this saying what the `std::abort()` is for?

benSepanski (approved):
Looks great to me! Had one small question

lib/VanguardSupport.h
Technius: Yeah, that's what the `std::abort()` is for (even though it's technically redundant).

PR #97, opened 2023-04-28
Circom tests now live here in Vanguard instead their own repo [VAN-223]

0xddom:
As we discussed I moved the tests here instead their own repo


0xddom:
CI ignored my tests because the compiler wasn't found even tho I added it to nix. @Technius can you check my nix script out? Maybe I did something wrong, dunno

nix/circom.nix
```diff
@@ -0,0 +1,37 @@
...
+    repo = "circom";
+    rev = "9ad63f2895a530e5bf83834491c89db631b01465";
+    hash = "sha256-HvqLzBbCeNWUHajAH3C6f7KmdXP7EPn1WUX5cTqcojA=";
+  };
+
+  cargoLock = {
+    lockFile = ./circom-Cargo.lock;
+    outputHashes = {
+      "inkwell-0.1.1" = "sha256-tPtwlpFEtuhM54lsdG+VDA9Z6bLfAHi6aJRWgenAVmg=";
+    };
```
Technius, line 24: This is probably not needed, unless circom also uses inkwell.

nix/circom.nix
```diff
@@ -0,0 +1,37 @@
...
+
+  src = fetchFromGitHub {
+    owner = "Veridise";
+    repo = "circom";
+    rev = "9ad63f2895a530e5bf83834491c89db631b01465";
+    hash = "sha256-HvqLzBbCeNWUHajAH3C6f7KmdXP7EPn1WUX5cTqcojA=";
+  };
+
+  cargoLock = {
+    lockFile = ./circom-Cargo.lock;
```
Technius, line 21: For circom2llvm, we need to provide a custom lockfile since it does not have one.

This might not be necessary for circom if it already has a lockfile.

nix/circom.nix
```diff
@@ -0,0 +1,37 @@
...
+    outputHashes = {
+      "inkwell-0.1.1" = "sha256-tPtwlpFEtuhM54lsdG+VDA9Z6bLfAHi6aJRWgenAVmg=";
+    };
+  };
+
+  postPatch = ''
+    cp ${./circom-Cargo.lock} Cargo.lock
```
Technius, line 28: This should be removed if the custom cargo lock is not needed.

nix/circom.nix
```diff
@@ -0,0 +1,37 @@
...
+  '';
+
+  buildInputs =
+    [ llvmPackages_13.libllvm libffi libiconv libxml2 zlib ]
```
Technius, line 32: This needs to be updated with whatever native dependencies are required by circom.

Technius (commented):
@0xddom You'll need to modify `nix/vanguard-driver.nix` so that it takes the circom package as a dependency. There's some code there for circom2llvm that you can follow.

Technius:
Ok, I think if there is a upstream Cargo.lock, we need to provide the sha256 hash of it. We can get it by setting the `cargoHash` to `lib.fakeHash`, trying to build again, and then using the "expected" hash in the error message.

0xddom:
All good now! Ready to merge

Technius (approved):

PR #98, opened 2023-04-28
Ben sepanski/support private repo [VAN-229]

benSepanski:
Sets up `Vanguard` to support private detector repositories, such as [circuitRecon](https://github.com/veridise/circuitRecon)


CMakeLists.txt
```diff
@@ -97,8 +97,42 @@ endif()
...
-install(EXPORT ${PROJECT_NAME}Targets DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME}")
+install(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/lib/"
+        DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}"
+        FILES_MATCHING PATTERN "**/*.h")
+
+# TODO: Prepend svf #includes in Vanguard with "svf/" so that we can
+# remove the forward slash here, installing a single "svf/" directory
+# instead of SVF-FE, Util, Graphs, CUDD, DDA, etc. all individually
+install(DIRECTORY "${SVF_INCLUDE_DIR}/"
+        DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}"
```
Technius, line 112: Why are we installing the SVF headers with CMake?

CMakeLists.txt
```diff
@@ -97,8 +97,42 @@ endif()
...
+        )
+
+install(EXPORT ${PROJECT_NAME}Targets
+        FILE ${PROJECT_NAME}Targets.cmake
+        NAMESPACE ${PROJECT_NAME}::
+        DESTINATION "${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME}"
+        )
```
Technius, line 119: I used the `${PROJECT_NAME}Targets` pattern earlier but that was before I discovered that it is an antipattern. It's actually better to use a fixed name to avoid backwards compatibility problems.

Note that this will change the name of the `*Targets.cmake` file.

lib/CMakeLists.txt
```diff
@@ -8,13 +8,14 @@ set(LLVM_LINK_COMPONENTS
   LLVMPasses
 )
 
-add_library(libVanguard)
```
Technius, line 11: Any particular reason why libVanguard was renamed to Vanguard?

lib/detectors/DetectorRegistry.h
```diff
@@ -58,13 +61,18 @@ namespace vanguard {
...
+        }
+
         Detector::DetectorDomain domain(const std::vector<std::string>& detectorNames);
 
         std::vector<std::string> all();
 
     private:
         static DetectorRegistry *instance;
         std::unordered_map<std::string, Detector::DetectorDomain> registry{};
+        std::vector<std::unique_ptr<ExternalDetectorRegistry>> externalRegistries;
```
Technius, line 75: Missing a `#include <vector>` at the top of the file

lib/Vanguard.h
```diff
@@ -0,0 +1,68 @@
...
+#include "domain/libBlockchain/SummaryReader.h"
+#include "domain/libCircuit/CircuitSummaryReader.h"
+#include "SARIF/Log.h"
+#include "SARIF/formats/JSONSerializer.h"
+#include "rapidjson/filewritestream.h"
+#include <rapidjson/writer.h>
+#include <rapidjson/ostreamwrapper.h>
+
+#include <memory>
+#include <utility>
```
Technius, line 56: Most of these includes are specific to the implementation and should be in the cpp file.

lib/CMakeLists.txt
```diff
@@ -8,13 +8,14 @@ set(LLVM_LINK_COMPONENTS
...
+target_link_libraries(Vanguard
   PUBLIC
   SVF::Svf
   ${LLVM_LINK_COMPONENTS}
 )
-target_include_directories(libVanguard PUBLIC
+target_include_directories(Vanguard PUBLIC
   "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>"
   "$<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>" #TODO: Change install interface?
 )
```
Technius, line 21: There is a missing

```cmake
install(TARGETS Vanguard
  EXPORT VanguardTargets
  LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
  ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
)
```

Technius: Oh, this is in the top-level CMakeLists.txt, but it should really belong in this file.

CMakeLists.txt
benSepanski: When including the vanguard headers in circuitRecon, I got a bunch of errors due to SVF not being there. I didn't want to add it as a dependency to circuitRecon since circuitRecon only depends on SVF indirectly

lib/CMakeLists.txt
benSepanski: The produced archive file was named `liblibVanguard.a` instead of `libVanguard.a`

Technius: I think that can be fixed with a call to `set_target_properties(libVanguard PROPERTIES OUTPUT_NAME "libVanguard${CMAKE_SHARED_LIBRARY_SUFFIX}")`, and it won't require a large renaming refactor.

CMakeLists.txt
Technius: Ugh, that's because SVF's targets are not set up correctly, and we're using `include_directory(...)` in the top-level CMakeLists.txt.

I don't really like this hack as it may cause undebuggable linker errors if SVF is not installed. I think it would be better to have SVF as an explicit dependency in circuitRecon. We can patch SVF to fix these issues later.

lib/CMakeLists.txt
benSepanski: I've undone the rename for now with a rebase.

The only minor ugly thing is now circuitRecon links against `vanguard::libVanguard` instead of `vanguard::Vanguard`, but that's fine

lib/detectors/DetectorRegistry.h
benSepanski: Added in https://github.com/Veridise/Vanguard/pull/98/commits/5c226150668a7cd549c6ef243908728b9542ad31

lib/Vanguard.h
benSepanski: Headers fixed/reorganized in https://github.com/Veridise/Vanguard/pull/98/commits/76e72fbd239146ac7f16dfc99ea6951c7539a0eb

lib/CMakeLists.txt
benSepanski: I'm just leaving it as lib lib for now, since CMake doesn't complain. I think I changed this because I thought it would fix another issue, but it wound up being a different problem in CMake

lib/CMakeLists.txt
benSepanski: Moved in https://github.com/Veridise/Vanguard/pull/98/commits/9dad935c5d86dbd3ef128530ca07d58648a062ce

CMakeLists.txt
benSepanski: Another reason I was hesitant to do this initially is because of all the patches and nix set up that was done to make it work in Vanguard

I'd like to avoid duplicating this if possible

Technius: I think you should be safe to use it in the recon project. SVF is already packaged in Vanguard, and since Vanguard is being consumed in recon, it should also be available via `find_package(SVF REQUIRED CONFIG)`.

lib/CMakeLists.txt
benSepanski: and https://github.com/Veridise/Vanguard/pull/98/commits/4cdd3d2a8d6aa620a1325147eac00fbcff78552c

CMakeLists.txt
Technius: Maybe we should discuss this synchronously. There are a lot of tiny issues that we'll need to fix in order to compose the two projects together.

benSepanski: Hmm, actually it still works if I just remove it. Thanks nix!

CMakeLists.txt
benSepanski: Removed in https://github.com/Veridise/Vanguard/pull/98/commits/f216255e4284a003b8376cb78dbd36ced71d2ec2

lib/detectors/DetectorRegistry.cpp
```diff
@@ -38,6 +38,14 @@ namespace vanguard {
             ReadWriteSet<BlockchainDomain>::domain());
         add(UncheckedReturnDetector<BlockchainDomain>::name(),
             UncheckedReturnDetector<BlockchainDomain>::domain());
+
+        // add externally registered detectors
+        for(const auto &externalRegistry : this->externalRegistries) {
+            auto nameToDomainMap = externalRegistry->getNameToDomainMap();
+            for(const auto &nameDomainPair : nameToDomainMap) {
+                add(nameDomainPair.first, nameDomainPair.second);
+            }
```
Technius, line 47: I don't think this will work. The external registries list will be empty in the constructor, and they can only be added after the larger `DetectorRegistry` is constructed.

benSepanski: Good catch, fixed in c0c77ce57ea3196322f4a8505d13acf46a786583

benSepanski:
@Technius Anything else on this PR you wanted touched up?

PR #102, opened 2023-05-05
Update flake.nix devShell to use propagatedBuildInputs

benSepanski:
Fix flake dev shell


flake.nix
```diff
@@ -33,7 +33,7 @@
 
     devShells = flake-utils.lib.flattenTree rec {
       default = pkgs.libVanguard.overrideAttrs (attrs: {
-        buildInputs = attrs.buildInputs
+        buildInputs = attrs.propagatedBuildInputs
```
Technius, line 36: This will "delete" the `buildInputs` on the `libVanguard` derivation if they exist. The original problem can be fixed with:
```suggestion
        buildInputs = (attrs.buildInputs or [])
```

Technius (changes requested):

benSepanski:
Thanks! I'm still kind of figuring out Nix. Made the suggested fix

Technius (approved):

PR #103, opened 2023-05-08
Improve Null Function Checking API

Sapo-Dorado:
Add `isImplemented()` function to `Function.h` that checks if the underlying llvm function is defined. This removes unnecessary calls to `unwrap()` within detectors to check if the underlying function is null


lib/program/Function.h
```diff
@@ -41,6 +41,13 @@ namespace vanguard{
             return function;
         }
 
+        bool isImplemented() const {
+            if(function) {
+                return true;
+            }
+            return false;
```
Technius, line 48:
```suggestion
            return static_cast<bool>(function);
```

Sapo-Dorado:
> This is a step in the right direction (hides LLVM implementation details), but I'm not sure this actually solves the root cause (methods on program module's Function will segfault if the LLVM function is not defined).
> 
> For example, it should be safe to call `name()` or `blocks()` without having to check `isImplemented()`. The implementation of `Function` itself should do the null checking so that the caller does not need to insert `if (function.isImplemented()) { ... }` everywhere.

That makes sense, how do you think we should handle the return value of `name()` and `blocks()`. Is it fine to return an empty string/empty vector? This might create weird outputs for detectors like the DebugSummary detector.

Technius:
For `blocks()`, I think it is fine to return a null iterator or a null vector or whatever. `name()` is a bit more tricky because that's something that actually needs to be read from the summary and stored in the `Function`. You make a good point about the empty name though, we should have some convention that specifies when the name is well-defined.

>  This might create weird outputs for detectors like the DebugSummary detector.

We'll want to update some of the detectors in response to the API updates anyways.

Sapo-Dorado:
@Technius Ok, I added the `IsImplemented()` checks to the Function api, so calls to the api don't seg fault and calling `IsImplemented()` from detectors is no longer required. Do you think this version is good to merge?

lib/program/Function.h
```diff
@@ -16,57 +16,82 @@ namespace vanguard{
...
+            }
+            return nullptr;
         }
 
         virtual bool hasBody() const  {
-            return !function->isDeclaration();
+            if(isImplemented()) {
+                return !function->isDeclaration();
+            }
+            return false;
```
Technius, line 48:
```suggestion
            return isImplemented() && !function->isDeclaration();
```

Technius (approved):

PR #104, opened 2023-05-08
A Basic Implementation of DBZ checker

wcphkust:
Hi @0xddom and @kferles 

I implemented a basic version of DBZ (divide-by-zero) checker. It can detect the DBZ bug in the following tiny program:

```
pragma circom 2.0.0;

template Edwards2Montgomery() {
  signal input in[2];
  signal output out[2];
  out[0] <-- (1 + in[1]) / (1 - in[1]);
  out[1] <-- out[0] / in[0];
  out[0] * (1-in[1]) === (1 + in[1]);
  out[1] * in[0] === out[0];
}

component main = Edwards2Montgomery();
```

Here is the output of the checker after running the command `Vanguard --vanguard-domain circuit --detectors dbz $CIRCOMIR_PATH/DBZ_testcase01.ll`:

```
starting
Running detector: dbz
        Collect Global Variables
[Source] AddrVFGNode ID: 21 AddrStmt: [Var88 <-- Var89] 
   %initial.in.input = alloca [2 x i128]*, align 8 
[Sink] BinaryOPVFGNode ID: 126 PAGEdge: [118 = Binary(117, 114, )]          %sub = sub i128 1, %read.in.input6 
[Sink] LoadVFGNode ID: 71 LoadStmt: [Var128 <-- Var127] 
   %read.in.input12 = load i128, i128* %array_getter11, align 4 
Analysing slice:21)

Guard computation for slice:21)
onReachable
=============== DBZ  REPORTS ===============
In component Edwards2Montgomery the following input signals can cause a dbz bug:
  - in
==========================================
```

I omit the intermediate log consoled by the taint engine because it is so verbose.

I am not sure whether the checker is fully implemented, as I might forget several subcases. If so, please inform me and I will polish the implementation. :)


lib/detectors/DBZ/DBZ.h
```diff
@@ -0,0 +1,188 @@
...
+
+        void detect(typename Domain::Universe &universe) override {
+            for (auto *unit : universe.units())
+                for (auto *component : unit->components()) {
+                    detectInComponent(component);
+                }
+        }
+
+        void report() override {
+            std::cout << "=============== DBZ  REPORTS ===============\n";
```
0xddom, line 148: This will need to be updated to the new reporting API, but I did not tell you about it so its ok for now.

lib/detectors/DBZ/DBZ.h
```diff
@@ -0,0 +1,188 @@
...
+            return name();
+        }
+
+        static Detector::DetectorDomain domain() {
+            return Detector::CIRCUIT;
+        }
+
+    private:
+        void detectInComponent(typename Domain::Component *component) {
+            this->runDBZAnalysis(component,
```
0xddom, line 178: nit: Is it necessary to use `this` here?

0xddom (changes requested):
Overall looks good for the reachability part. Lets take the ideas here for the graphs as we discussed earlier and then we can merge as a first iteration of the detector.

lib/detectors/DBZ/DBZ.h
```diff
@@ -0,0 +1,188 @@
...
+            ValueVec outputSignals;
+            for (auto *signal : component->outputSignals()) {
+                outputSignals.push_back(signal->unwrap());
+            }
+            return outputSignals;
+        }
+
+        ValueVec getDivisorSignals(typename Domain::Component *component) {
+            ValueVec divisorSignals;
+            for (auto& BB : *(component->getLogic()->unwrap())) {
```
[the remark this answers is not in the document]
kferles: To add to that, this part shouldn't even depend on LLVM's API. Don't our wrappers have a similar API?

kferles (changes requested):
Only added to the conversation started by Daniel. If our current API can handle the pattern matching, we should prefer it over LLVM's API.

lib/detectors/DBZ/DBZ.h
wcphkust: I have fixed it.

wcphkust:
I have removed the LLVM API for the iteration. However, after the refactoring, the detector fails to report the bug. I could not figure out which parts are not equivalent in the refactoring.

kferles:
@0xddom: Can you help @wcphkust with the Vanguard issue mentioned above?

Technius:
Please add a lit test that checks whether the detector is working correctly. In addition to helping debug, it also helps catch errors that might occur on different machines. See examples here: https://github.com/Veridise/Vanguard/tree/v1/test/circom

lib/detectors/DBZ/DBZ.h
```diff
@@ -47,6 +47,19 @@ namespace vanguard {
         using Val2Signal = std::map<const llvm::Value *, typename Domain::template Signal<llvm::Value> *>;
         using SignalVec = std::vector<typename Domain::template Signal<llvm::Value> *>;
 
+        /**
+        * Finds all the Binary Operators in a function
+        */
+        class BinaryOpQuery : public InstructionClassQuery<Domain, const llvm::Value *> {
+        public:
+            virtual void visit(const UnknownIns <Domain> &v) {
```
0xddom, line 55:
```suggestion
            virtual void visit(const BinaryOpIns <Domain> &v) {
```

There are several classes that wrap instructions used depending on the subjacent llvm instruction. For the instructions you want I think you need to use `BinaryOpIns`

lib/detectors/DBZ/DBZ.h
```diff
@@ -0,0 +1,202 @@
...
+            return outputSignals;
+        }
+
+        ValueVec getDivisorSignals(typename Domain::Component *component) {
+            ValueVec divisorSignals;
+            BinaryOpQuery q;
+            std::vector<const llvm::Value *> values = component->getLogic()->findAll(q);
+
+            for (int i = 0; i < values.size(); i++) {
+                if (auto bop = llvm::dyn_cast<llvm::BinaryOperator>(values.at(i))) {
```
0xddom, line 102: You could also put this statement inside the query class and make it only return `IntDivRem` instructions

wcphkust: OK. I have polished it. The current implementation can detect the bug correctly.

0xddom:
Ok, good! I think the Constraints and Dataflow graph implementations can be made in another PR and we merge this one after we add tests to it. @kferles what do you think?

kferles:
Yes, definitely! Let's do separate PRs for those.

wcphkust:
I have removed `NDD.h` but I do not know how to add a lit test to check this detector

0xddom:
The reporting lgtm

0xddom (approved):

kferles (approved):

PR #105, opened 2023-05-12
A Basic Implementation of NDD checker

wcphkust:
Hi @kferles and @0xddom,

I implemented a basic version of NDD (Non-deterministic dataflow) checker. It can detect the NDD bug in the following tiny program:

```
pragma circom 2.0.0;

template IsZero() {
    signal input in1;
    signal input in2;
    signal output out;
    signal inv;
    inv <-- in1!=0 ? 1/in1 : 0;
    in2 === 2;
    out <== -in1*inv +1;
    in1*out === 0;
}

component main {public [in1, in2]}= IsZero();
```

Here is the output of the checker after running the command `Vanguard --vanguard-domain circuit --detectors ndd $CIRCOMIR_PATH/NDD_testcase01.ll`:

```
========================
 1 NDD signal detected: 
========================
Potential NDDs in IsZero are caused by: 
--------------------------------------------------------------------------------
  * Signal: in1
--------------------------------------------------------------------------------
```

The above simple test case has been uploaded to the test directory. I also simplified the implementation of the DBZ checker and abstracted several commonly used methods, e.g., `inputValues` and `outputValues`, into the class component. Hope such refactoring could improve the cleanliness and maintainability.


wcphkust:
Another proposal is that we may need to create a new subclass of `UniverseDetector`. Currently. the DBZ and NDD detectors are essentially source-must-not-sink value-flow bug detectors. I think we need to create at least two kinds of Detector, namely  `source-must-not-sink` and `source-always-sink`. Then the taint analysis engine is used as the search engine over the graph, while the two kinds of basic bug detectors support two lines of instantiations of detection.

The proposal is motivated by improving the abstraction level of the detectors. Current implementations of NDD and DBZ detectors share the same detection mode although the duplicated/similar code of the detection does not take up a large proportion of the whole source code. I am posting the proposal here for more discussion.

lib/domain/libCircuit/Component.h
```diff
@@ -52,6 +52,22 @@ namespace vanguard {
             return this->signalsOfType(OUTPUT_SIGNAL);
         }
 
+        [[nodiscard]] std::vector<const llvm::Value *> inputValues() const {
```
kferles, line 55: I am not sure whether this is the correct place for these methods. Maybe @0xddom can comment on that. But, generally we want to keep

lib/detectors/NDD/NDD.h
```diff
@@ -0,0 +1,145 @@
...
+        /**
+         * Finds all the call instructions performing switch
+         */
+        class SwitchCallQuery : public InstructionClassQuery<Domain, const llvm::CallInst*> {
+        public:
+            virtual void visit(const CallIns<Domain> &v) {
+                auto *fun = (llvm::CallInst*) v.unwrap();
+                llvm::Function *callee = fun->getCalledFunction();
+                if (callee != nullptr) {
+                    if (callee->getName().str().find("fn_intrinsic_utils_switch") != std::string::npos) {
```
kferles, line 58: @0xddom: Will this be the same in the new frontend?

0xddom: No, it will not. We should move `SwitchCallQuery` to `lib/domain/libCircuit/langs/circom/CircomSearchers.h` since its logic is specific of `circom2llvm`

lib/domain/libCircuit/Component.h
0xddom: And of course mode the function to somewhere that makes sense like a static method of `Variable` for example or a function in `ValueClasses.h` or `Value.h`

0xddom:
Wrt the detection logic itself lgtm so far

lib/domain/libCircuit/Component.h
wcphkust: Good idea. I will refactor it in this way.

lib/program/Value.h
```diff
@@ -13,6 +13,15 @@
 
 namespace vanguard{
 
+    template <typename T>
+    std::vector<const llvm::Value *> unwrapMany(std::vector<T> vars) {
+        std::vector<const llvm::Value *> values;
+        for (auto *signal : vars) {
```
0xddom, line 19: nit: I would name them `wrappers` instead of `vars` and `wrapper` instead `signal`. Since the only thing we care about of those classes is that they implement the `unwrap()` method.

0xddom:
> Another proposal is that we may need to create a new subclass of `UniverseDetector`. Currently. the DBZ and NDD detectors are essentially source-must-not-sink value-flow bug detectors. I think we need to create at least two kinds of Detector, namely `source-must-not-sink` and `source-always-sink`. Then the taint analysis engine is used as the search engine over the graph, while the two kinds of basic bug detectors support two lines of instantiations of detection based on the taint analysis engine.
> 
> The proposal is motivated by improving the abstraction level of the detectors. Current implementations of NDD and DBZ detectors share the same detection mode although the duplicated/similar code of the detection does not take up a large proportion of the whole source code. I am posting the proposal here for more discussion.

I can't remember if I answered this already in another channel but here are my thoughts anyway.
I think those abstractions are useful since they seem to be common patterns tho I'm not sure they should be encoded in a detector itself. The way I understand the detectors (@kferles or @stephensj2 correct me if I'm wrong) is that they should be as idiomatic as possible to help explain the thing they are detecting with pesky details abstracted away in analysis modules.

That is also why I think modules like the dataflow and constraint graph abstractions could be useful as well (and can be built on top of the abstractions you propose) because it would allow us to write code that reads closer to a description of the bug.

Another consideration is that by encoding the abstraction in the inheritance tree we are locking the detector to a class of algorithms when the same problem may be solved with different techniques. Imagine for a particular detector we switch from reachability with SVF to, for example, abstract interpretation we would need to basically rewrite the detector from scratch whereas with modules we just replace one module with another keeping everything else in the detector intact.

tl;dr: I like the idea of creating a couple classes that represent `source-must-not-sink` and `source-always-sink` but not as detectors

lib/program/Value.h
wcphkust: I fixed it.

0xddom (approved):
Nice, merge whenever you are ready

PR #107, opened 2023-05-12
Ben sepanski/refactor detectors

benSepanski:
This PR removes the out-dated Reentrancy detector, and switches DebugSummary to use the new report formatter API.

This is the first of several PRs updating stable detectors to the new report formatter API. Since the various detectors have different authors, and I want to keep the review sizes manageable, I'll be making several PRs for this tasks.


lib/detectors/Debug/DebugSummaryDetector.h
```diff
@@ -137,8 +163,14 @@ namespace vanguard {
...
+            formatSettings.list.extraNewlineDepth = 0;
+            formatting::PlainTextFormat plainTextFormat;
+            plainTextFormat.setSettings(formatSettings);
+
+            auto report = info.getReport();
+            llvm::outs() << plainTextFormat.format(report);
+            llvm::outs() << "Done!\n";
         }
 
         static std::string name() {
```
Sapo-Dorado, line 176: Should we also update the name of the detector to adhere to the new naming convention (debug-summary)? The changes look good to me otherwise.

benSepanski: I think that's a good idea, addressed in https://github.com/Veridise/Vanguard/pull/107/commits/c55743f5bfe368b43f4c1e7a7efac2ce132e36ba.

@Technius would this need to be updated in SaaS?

kferles (approved):
LGTM

lib/detectors/Debug/DebugSummaryDetector.h
Technius: It's not used there right now.

PR #108, opened 2023-05-15
Add new tests for the circom compiler frontend

0xddom:
New tests for the circom compiler and some reorganization. Last week I got things mixed up by mistake and I dropped them in the wrong repo. Fixing this now. Includes the tests @tim-hoffman added to the other repo. They may need to be fixed wrt lit and CI


kferles (changes requested):
Looks like a trivial PR. Although, CI is failing. Also, didn't we decide to move these tests to the frontend repo?

In any case, I am fine with these tests living here for now. But let's make sure CI is passing before merging.

0xddom:
@kferles finally fixed. Approve when you consider plz

PR #110, opened 2023-05-25
A Basic Implementation of DCD checker

wcphkust:
Hi @kferles and @0xddom,

I implemented a basic version of the DCD (Dataflow-Constraint Discrepancy) checker. The design of the DCD checker consists of the following components.

- Compute the data-flow reachability over all the values based on the taint analysis engine.
- Compute the data-flow reachability between the values and LHS/RHS values of constraints. For each constraint, we compute the constraint reachability of the values reaching the LHS/RHS.
- Considering the load/store instructions, convert the data-flow/constraint values to the addresses (of input/output) storing them.
- Diff the data-flow reachability and constraint reachability over input/output values, which yield a relation indicating the dataflow-constraint discrepancy.

In the implementation, the third component is fused into the first two parts, which correspond to the method `runDFRAnalysis` and `runCRAnalysis`, respectively.

Based on the above design, Vanguard can detect the DCD bug in the following tiny program:

```
pragma circom 2.0.0;

template SingleAssignment0() {
  signal input a;
  signal input b;
  signal output out;
  out <-- a + 1;
  out === b + 1;
}
component main = SingleAssignment0();
```

Here is the output of the checker after running the command `Vanguard --vanguard-domain circuit --detectors dcd $CIRCOMIR_PATH/dcd_buggy.ll`:

```
========================
 1 DCD signal detected: 
========================
A potential DCD in SingleAssignment0 is caused by: 
--------------------------------------------------------------------------------
  * Signals: initial.a.input and initial.out.output
--------------------------------------------------------------------------------
```

The above simple test case has been uploaded to the test directory.

Remark: Currently our implementation cannot support array signals. Given current IR as input, I think it is not difficult to polish the detector to support them. I have not started this part as @0xddom said the IR translator will be upgraded and the input form might be changed.


lib/detectors/DCD/DCD.h
```diff
@@ -0,0 +1,327 @@
...
+
+            //Lift to input/output values based on sigFlowReachResult
+            if (sigFlowReachResult.find(component) != sigFlowReachResult.end()) {
+                for (auto it : sigFlowReachResult[component]) {
+                    const llvm::Value* value1 = it.first;
+                    for (const llvm::Value* value2 : it.second) {
+                        const llvm::Value* signalValue1 = nullptr;
+                        const llvm::Value* signalValue2 = nullptr;
+
+                        if (std::find(inputValues.begin(), inputValues.end(), value1) != inputValues.end() ||
```
0xddom, line 74: This block is duplicated below. Extract it into a function to avoid duplication. Same with the conditionals that use `std::find`. Extracting those predicates will make the code more readable. I drafted a couple example implementations here https://godbolt.org/z/xfMMEEGEr

lib/detectors/DCD/DCD.h
```diff
@@ -0,0 +1,327 @@
...
+                                for (const llvm::Value * value4 : storeValueToAddrMap[value2]) {
+                                    if (std::find(inputValues.begin(), inputValues.end(), value4) != inputValues.end() ||
+                                        std::find(outputValues.begin(), outputValues.end(), value4) != outputValues.end()) {
+                                        signalValue2 = value4;
+                                    }
+                                }
+                            }
+                        }
+
+                        if (signalValue1 != nullptr && signalValue2 != nullptr) {
```
0xddom, line 102:
```suggestion
                        if (signalValue1 && signalValue2) {
```

lib/detectors/DCD/DCD.h
```diff
@@ -0,0 +1,327 @@
...
+
+                    TaintAnalysisResultsCollector RHSCollector;
+                    TaintAnalysis analysis2(focusValues, signalRightSinkValues, &RHSCollector);
+                    analysis2.run();
+
+                    for (const auto& [fromValue, valueFlows] : RHSCollector) {
+                        for (const auto& toValue : valueFlows.sinks) {
+                            sigConRightReachResult[component][callInst].emplace(fromValue);
+                        }
+                    }
```
0xddom, line 146: Duplicate code here too. Can be abstracted in a function.

lib/detectors/DCD/DCD.h
```diff
@@ -0,0 +1,327 @@
...
+                        } else {
+                            if (loadValueToAddrMap.find(value1) != loadValueToAddrMap.end()) {
+                                for (const llvm::Value * value3 : loadValueToAddrMap[value1]) {
+                                    if (std::find(inputValues.begin(), inputValues.end(), value3) != inputValues.end() ||
+                                        std::find(outputValues.begin(), outputValues.end(), value3) != outputValues.end()) {
+                                        signalValue1 = value3;
+                                    }
+                                }
+                            }
+                        }
```
0xddom, line 178: And here we have the same pattern as in `runDFRAnalysis`. More reasons to extract this block in a function

lib/detectors/DCD/DCD.h
```diff
@@ -0,0 +1,327 @@
...
+                                for (const llvm::Value * value4 : storeValueToAddrMap[value2]) {
+                                    if (std::find(inputValues.begin(), inputValues.end(), value4) != inputValues.end() ||
+                                        std::find(outputValues.begin(), outputValues.end(), value4) != outputValues.end()) {
+                                        signalValue2 = value4;
+                                    }
+                                }
+                            }
+                        }
+
+                        if (signalValue1 != nullptr && signalValue2 != nullptr) {
```
0xddom, line 194:
```suggestion
                        if (signalValue1 && signalValue2) {
```

lib/detectors/DCD/DCD.h
```diff
@@ -0,0 +1,327 @@
...
+        };
+
+        using ValueVec = std::vector<const llvm::Value *>;
+        using Val2Signal = std::map<const llvm::Value *, typename Domain::template Signal<llvm::Value> *>;
+        using ValConReachability = std::map<const llvm::CallInst *, std::unordered_set<const llvm::Value*>>;
+        using ValReachability = std::map<const llvm::Value *, std::unordered_set<const llvm::Value*>>;
+
+    protected:
+        //Data-flow reachability analysis
+        void runDFRAnalysis(typename Domain::Component *component, ValueVec focusValues, ValueVec inputValues, ValueVec outputValues) {
```
0xddom, line 54: This function should be extracted in its own class as an analysis for reusability.

`void runDFRAnalysis(...)` -> `class DRFAnalysis { void run(...) }`

lib/detectors/DCD/DCD.h
```diff
@@ -0,0 +1,327 @@
...
+                        if (signalValue1 != nullptr && signalValue2 != nullptr) {
+                            sigDFReachResult[component][signalValue1].emplace(signalValue2);
+                        }
+                    }
+                }
+            }
+        }
+
+        //Constraint reachability analysis
+        void runCRAnalysis(typename Domain::Component *component, ValueVec focusValues, ValueVec inputValues, ValueVec outputValues) {
```
0xddom, line 111: This function should be extracted to a class as well.

0xddom (changes requested):

lib/detectors/DCD/DCD.h
```diff
@@ -0,0 +1,327 @@
...
+
+        [[nodiscard]] std::string getName() const override {
+            return name();
+        }
+
+        static Detector::DetectorDomain domain() {
+            return Detector::CIRCUIT;
+        }
+
+        void dumpReachability(typename Domain::Component *component) {
```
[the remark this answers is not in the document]
wcphkust: It is just debugging. But I think it would be useful to help understand the reachability relation between the signals.

lib/detectors/DCD/DCD.h
wcphkust: I have refactored all the other parts except this one. I think it is a good practice to avoid duplication, while too many abstractions with such a short function would make us define many helper functions in various modules. Although the code line is a bit long, I don't think the code is hard to understand. Personally speaking, the functions of a class should explicitly reflect the key operations/functionalities. Over-abstraction is not a good choice in some cases, I think.

0xddom (approved):
LGTM

